PR #485, opened 2020-01-10
feat: typescript

anucreative:


remorses:
Does this means this library will have types by default?

The only thing that prevents me from using these components is typescript support, i am so happy that you will add them directly in the packages

Awesome work!

trustin:
Thanks a lot for working on this pull request and the project itself. Could you let us know the status of this pull request? I'm curious about when this will land or what the plan around this is. Given the number of components to migrate, it may be better migrating incrementally.

anucreative:
Hi @remorses, @trustin 👋 

Yes, the goal is to move everything to Typescript but it may take us a while. And, yes, we'll probably do it incrementally. This PR was the result of a few hours work to see how feasible it was. We'll keep you up-to-date on progress but I expect the next opportunity we'll have to work on this (Typescript migration) will be next month.

theo-mesnil:
Old PR, typescript is not planned ofr the moment :)

theo-mesnil:
https://github.com/WTTJ/welcome-ui/pull/854

trustin:
Excellent work, folks! :heart:

PR #517, opened 2020-01-30
chore: add component README files

anucreative:


theo-mesnil:
Maybe it's better to call entry documentation -> homepage to be compatible with package json and npm website :)

PR #545, opened 2020-03-02
fix: card cover style props not being sent to component

P1X3L:


package.json
```diff
@@ -104,6 +104,7 @@
     "lerna": "^3.19.0",
     "lerna-update-wizard": "^0.17.7",
     "lodash.capitalize": "=4.2.1",
+    "lodash.difference": "^4.5.0",
```
anucreative: We shouldn’t need lodash.difference...

P1X3L: Ok I'm removing it!

PR #562, opened 2020-03-25
fix: cursor text on search field

guillaumewttj:


packages/Search/styles.js
```diff
@@ -27,7 +27,7 @@ export const Input = styled(filterFieldComponent('input', FILTER_PROPS))(
       css`
         padding-left: ${th(`fields.sizes.${size}.height`)};
       `};
-    cursor: default;
+    cursor: text;
```
theo-mesnil: and if you delete this it's work ?

guillaumewttj: yep it works, my bad !

PR #570, opened 2020-04-04
chore(docz): upgrade to 2.3.1

theo-mesnil:
wait for an answer from docz, headings are not been applied with v2.3.0.


theo-mesnil:
https://github.com/doczjs/docz/issues/1443

PR #575, opened 2020-04-06
fix: datePicker infinite loop due to bad date parse into useEffect

guillaumewttj:


guillaumewttj:
[Members#form : crash while using the birthday datetimepicker](https://trello.com/c/arZCNX2t/175-membersform-crash-while-using-the-birthday-datetimepicker)

PR #583, opened 2020-04-08
feat: automate icon font

anucreative:
- Main icon script is easier to read 😝
- Re-generates unicode.json if new icons found in `icons/_assets`
- Re-generates font and saves to `packages/IconFont/fonts`

~~Note: some icons break the script at the moment:~~
~~- Those with transforms (e.g. the flags) so have to recreate these without transforms~~
~~- Some other icons that appear to be the new ones (e.g. clock, salary) so have to debug these~~

Questions:
- Do we need the script to generate the css or is this already taken care of with `packages/IconFont/styles.js`?
- Should we be hosting the fonts at `cdn.welcometothejungle.com` or should users include it from wui (if so, how)?


scripts/icons.js
```diff
@@ -19,50 +21,110 @@ const toPascalCase = str => {
...
+          const [key, type] = file.split('.')
+          if (type === 'svg') {
+            return fs
+              .readFileAsync(path.join(inputPath, file), 'utf8')
+              .then(content => ({ key, content }))
+          }
+        })
+      )
+    )
+    .then(files => files.filter(Boolean))
```
Calvein, line 40: On pourrait filtrer sur les fichiers qui sont de type svg [avant le map là](https://github.com/WTTJ/welcome-ui/pull/583/files#diff-14430e385ad8a0f7f7aff14f103aef03R30) plutôt non ?

Calvein:
> * Should we be hosting the fonts at `cdn.welcometothejungle.com` or should users include it from wui (if so, how)?

Why not both? Be if we have to choose, I'd rather go with the CDN.

anucreative:
> > * Should we be hosting the fonts at `cdn.welcometothejungle.com` or should users include it from wui (if so, how)?
> 
> Why not both? Be if we have to choose, I'd rather go with the CDN.

My worry is just around publishing the file to the CDN when we publish a new version of `IconFont` and wanting to keep everything together. Always open though. And I think including a font in the HTML directly is still more common practise than importing it in a React component…

PR #595, opened 2020-04-10
fix(tabs): tab panel should use tabId prop instead of id

diegohaz:
On the Reakit's `TabPanel` component, `stopId` has been renamed to `tabId`, not `id`.

Note also that both `Tab`'s `id` and `TabPanel`'s `tabId` aren't required anymore. Reakit now uses the DOM order to determine the `Tab`/`TabPanel` relationship, although you can still set `id`/`tabId`, especially if you render panels out of order.

### DIFF
--- a/packages/Tabs/doc.mdx
+++ b/packages/Tabs/doc.mdx
@@ -47,19 +47,19 @@ Set the tab activated by default using `selectedId` on `useTabState`.
             Tab 5
           </Tab>
         </Tab.List>
-        <Tab.Panel {...tab} id="tab1">
+        <Tab.Panel {...tab} tabId="tab1">
           Tab.Panel 1
         </Tab.Panel>
-        <Tab.Panel {...tab} id="tab2">
+        <Tab.Panel {...tab} tabId="tab2">
           Tab.Panel 2
         </Tab.Panel>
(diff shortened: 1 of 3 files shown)


theo-mesnil:
@diegohaz thanks for you PR 🤩

I use id because in your documentation is `id` to set selected tab https://reakit.io/docs/tab/#default-selected-tab, maybe is wrong on the documentation ?

Do you use our design system personally ? :)

diegohaz:
On the `Tab` component it's `id` because it effectively set the `id` attribute on the tab element. But on `TabPanel` it's `tabId` because it represents the tab reference, not its own `id`.

You can still set `id` on `TabPanel`, and it'll set the `id` attribute on the panel element, but it has nothing to do with the relationship between the tab and its panel.

The example on master works because you don't really need to pass `tabId` to `TabPanel`. It'll infer the corresponding tab by the DOM order. If you inspect the HTML output, you'll see that `Tab` and its `TabPanel` have the same `id` attribute, which is an invalid markup.

I'm not using Welcome UI in any serious project yet, just playing with it. Thank you for the awesome work! :)

theo-mesnil:
@diegohaz ok i understand now :) thank you !

PR #597, opened 2020-04-14
Feat loader

guillaumewttj:
This loader look like the loader into this button : https://codepen.io/eriscolaro/pen/RwWbMVM


packages/Loader/index.js
```diff
@@ -0,0 +1,19 @@
...
+    <S.LoadingDot backgroundColor={color} shape="circle" size={size} />
+    <S.LoadingDot backgroundColor={color} shape="circle" size={size} />
+  </S.LoaderWrapper>
+))
+
+Loader.displayName = 'Loader'
+
+Loader.propTypes = {
+  color: string.isRequired,
+  size: oneOf('sm', 'md', 'lg')
```
theo-mesnil, line 18:
```suggestion
  size: oneOf(['sm', 'md', 'lg'])
```

packages/Loader/styles.js
```diff
@@ -0,0 +1,41 @@
...
+  ({ size }) =>
+    css`
+      width: ${th(`loaders.sizes.${size}`)};
+      height: ${th(`loaders.sizes.${size}`)};
+    `
+)
+
+export const LoaderWrapper = styled.div`
+  display: flex;
+  > ${LoadingDot} {
```
theo-mesnil, line 26: line break, c'est obligé le `>` ?

packages/Loader/index.js
```diff
@@ -0,0 +1,19 @@
+import React, { forwardRef } from 'react'
+import { oneOf, string } from 'prop-types'
+
+import * as S from './styles'
+
+export const Loader = forwardRef(({ color, dataTestId, size = 'sm', ...rest }, ref) => (
+  <S.LoaderWrapper data-testid={dataTestId} ref={ref} {...rest}>
+    <S.LoadingDot backgroundColor={color} shape="circle" size={size} />
```
theo-mesnil, line 8: add a color by default on theme, and add the possibilty to pass a width and height props if we want to put another size custom ?

packages/Loader/styles.js
```diff
@@ -0,0 +1,41 @@
...
+export const LoadingDot = styled(Shape)(
+  ({ size }) =>
+    css`
+      width: ${th(`loaders.sizes.${size}`)};
+      height: ${th(`loaders.sizes.${size}`)};
+    `
+)
+
+export const LoaderWrapper = styled.div`
+  display: flex;
```
theo-mesnil, line 25:
```suggestion
  display: flex;
  ${system};
```

packages/Loader/doc.mdx
```diff
@@ -0,0 +1,38 @@
...
+<TagVersion version={version} name={name} />
+
+## Install and import
+
+<Usage component={component} name={name} peerDependencies={peerDependencies} />
+
+## Examples
+
+<Playground>
+  <Loader color="black" />
```
theo-mesnil, line 23: remove color here, and add an exemple with color property

packages/Loader/styles.js
Calvein: Je rejoins Théo, à priori il n'y a pas besoin de nester `LoadingDot` dans `LoadingWrapper`

packages/Loader/index.js
guillaumewttj: the size is configured with the theme so you can create your own width/height for any size

theo-mesnil: yes i know, but we need to do the same everywhere and for exemple an icon can have a size prop or width and height for special case

PR #598, opened 2020-04-14
feat: add Avatar component

### LINKED ISSUE #596: Avatar component
### Motivation

Create an avatar component with an image, or if not a background color and initials.

theo-mesnil:
Fixes #596 

Capture d’écran 2020-04-14 à 11 48 54


packages/Avatar/index.js
```diff
@@ -0,0 +1,61 @@
...
+}) {
+  const theme = useTheme()
+  const backgroundColor = color || randomColor(theme && theme.colors && theme.colors.sub)
+  const avatarSize = theme.avatars.sizes[size]
+  const avatarFontSize = fontSize || `calc(${width ? theme.toRem(width) : avatarSize} / 2.5)`
+
+  function getInitials() {
+    const initials = words[1] ? `${words[0][0]}${words[1][0]}` : words[0][0]
+    return initials && initials.toUpperCase()
+  }
```
Calvein, line 28: Could be defined out of the component (or better, in `utils` like `randomColor` is)

packages/Avatar/utils.js
```diff
@@ -0,0 +1,5 @@
+export function randomColor(colors) {
```
Calvein, line 1: it'd be great if `randomColor` could take a seed argument that would generate the index rather than having a random one on each refresh

Calvein: For instance (I changed the name of the function because it's not random anymore):
```js
export function getColor(colors, words) {
  const colorKeys = Object.keys(colors)
  const index = words.join('').length % colorKeys.length
  return colors[colorKeys.find((color, i) => i === index)]
}
```

anucreative: I like this. 😻

Last line could just be ?

return colors[colorKeys[index]]

theo-mesnil: It's done @Calvein @anucreative  :) thanks 🏩

theo-mesnil:
@anucreative i changed the api to get initials :) thx !

packages/Avatar/index.js
theo-mesnil: Done :)

PR #641, opened 2020-06-01
Fix search warnings

guillaumewttj:
fixed:

- Controlled to uncrontrolled react warning when returning undefined value from itemToString
- Stylelint error into loader (breaks the lint test on master)


packages/Search/index.js
```diff
@@ -90,10 +90,13 @@ export const Search = forwardRef(
       setResults([])
     }
 
+    // prevents controlled to uncontrolled switch when itemToString returns falsy value
+    const handleItemToString = item => itemToString(item) || ''
```
anucreative, line 94:
```suggestion
    const handleItemToString = item => itemToString(item) ?? ''
```

Pour prevenir les items ou `itemToString(item)` === 0

PR #648, opened 2020-06-08
fix: add new loader size

guillaumewttj:


theo-mesnil:
Missing documentation

PR #747, opened 2020-07-21
feat: remove wrapping div added when using `shouldHideFocusRingOnClick`

P1X3L:

### DIFF
--- a/docs/pages/getting-started.mdx
+++ b/docs/pages/getting-started.mdx
@@ -55,7 +55,7 @@ export default function Root() {
 
 ## Hide the focus ring on mouse move or on keydown
 
-WuiProvider takes a `shouldHideFocusRingOnClick` prop which defaults to `true`. It hides the focus ring on mouse move or on keydown.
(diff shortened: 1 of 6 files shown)


docs/pages/getting-started.mdx
```diff
@@ -46,11 +46,29 @@ export default function Root() {
...
     </WuiProvider>
   )
 }
 ```
+
+## Hide the focus ring on mouse move or on keydown
+
+WuiProvider takes a `shouldHideFocusRingOnClick` prop which defaults to `true`. It hides the focus ring on mouse move or on keydown.
+
+It only adds styles to remove the outline on focus. If you need to remove another css property, you have to write your own styles using the `data-wui-hidefocusrings` data-attribute.
```
Calvein, line 60:
```suggestion
It only adds styles to remove the outline on focus. If you need to remove another css property, you have to write your own styles using the data-attribute `hideFocusRingsDataAttribute` exported from `@welcome-ui/utils`.
```

PR #750, opened 2020-07-23
fix: logic when numberOfSlides isn't divisible by slidesToSwipe

anucreative:
Logic wasn't quite right when showing an even number of slides in a carousel with an odd number of slides.

https://www.loom.com/share/db88f37b581c4f6eaeaf2ef09f9569e7

### DIFF
--- a/packages/Swiper/index.js
+++ b/packages/Swiper/index.js
@@ -34,11 +34,9 @@ export const Swiper = forwardRef((props, ref) => {
   }, [children, setNumberOfSlides])
 
   // Get array with indexes of visible slides so we know which ones are (aria-)hidden
-  const isVisible = slide =>
-    Array(slidesToShow)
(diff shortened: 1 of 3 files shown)


packages/Swiper/index.js
```diff
@@ -33,6 +33,13 @@ export const Swiper = forwardRef((props, ref) => {
     setNumberOfSlides(children.length || children.size)
   }, [children, setNumberOfSlides])
 
+  // Get array with indexes of visible slides so we know which ones are (aria-)hidden
+  const isVisible = slide =>
+    Array(slidesToShow)
+      .fill()
+      .map((_, idx) => pageIdx + idx)
```
Calvein, line 40: I reckon this should be outside `isVisible` since it doesn't depend on the args

PR #763, opened 2020-08-06
Adjust link to illustration on README

iaurg:
The link is broken, and I think that's the correct new path. :tada:


theo-mesnil:
Thanks for your contribution :)

PR #793, opened 2020-10-16
fix: updating icon font

anucreative:
We were recreating the unicode list each time we built the font which meant that icons were being misplaced e.g. 

If I have…
```
  "bold": "\f10f",
  "bulb": "\f110",
  "bulk": "\f111",
```

and I add a `book` icon I get…
```
  "bold": "\f10f",
  "book": "\f110",
  "bulb": "\f111",
  "bulk": "\f112",
```

where the `bulb` and `bulk` icons (and everything after) get displaced. 

This would mean that we'd have to version the font (and keep all old versions in the CDN) and it complicates the deployment process.

By adding new fonts to the end of the unicode list, we can continue to publish the font to our CDN and users with the old unicode list will still get the correct icons even with the new font.

We've reverted to the last stable unicode list (from before #774) and re-added the new icons from the previous commits (#782, #786, #787)


guillaumewttj (approved):

mleralec (approved):

theo-mesnil (approved):
🔥

PR #796, opened 2020-10-21
Add render props to FileUpload (and fixes)

Calvein:
I need to be able to replace the Button set in the render prop with a component to prevent the page to jump, so I added `file` & `onRemoveFile` to the render prop.
In addition, I've fixed an issue with `handleRemove` that was taking an event but was expected an index, which meant we were possibly removing the wrong one when multiple files were uploaded.
And last (and the one I'm really keen to get a review of) is that I added a key prop on the input based on the current files, to allow React to re-render this input thus allowing the user to upload a file, removing it & then re-adding it. I feel that it was needed because we can't manipulate the value of an input[type=file] element for security reasons.

https://www.loom.com/share/a35b315b60694936a63128d7567dbad1

### DIFF
--- a/docs/pages/fields/file-upload.mdx
+++ b/docs/pages/fields/file-upload.mdx
@@ -172,6 +172,28 @@ function() {
 }
 ```
 
+You can also set the `preview` prop to `null` and handle your own preview in place of the upload button with `files` & `onRemoveFile` render props:
+
+```jsx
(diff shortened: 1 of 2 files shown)


mleralec (approved):
👏

anucreative (approved):
Very nice, sir. 🎩

PR #809, opened 2020-12-21
fix: icon font generation

anucreative:
Needs more investigation (but it's the holidays). Converting to `.woff` seems to work but not `.woff2`. Du coup, switching to use `.woff` for icon font temporarily


theo-mesnil (approved):

guillaumewttj (approved):

anucreative:
Replaced by https://github.com/WTTJ/welcome-ui/pull/820

PR #822, opened 2021-01-20
chore: don't launch release task if `ssh-add -L` fails

P1X3L:
If your ssh key is not correctly set, the `yarn release` task will fail when trying to push new tags and exit in an unstable state (newly edited npm packages not published).

To avoid this, this proposal just tests that your ssh-key is correctly set before launching the release. If not, you get a warning.


theo-mesnil (approved):
Maybe you can add also this on prerelease ? :)

PR #837, opened 2021-03-25
fix: break all words on Text component with lines set to 1

P1X3L:
`word-break` is set to `break-all` when you use `Text` component with `lines` prop set to `1`


theo-mesnil (approved):
🔥

Calvein (approved):

guillaumewttj (approved):

PR #846, opened 2021-05-03
feat: use WUI cdn for emojis

Calvein:
We'll monitor the cost of the CDN and we might use an external CDN if it's too expensive


theo-mesnil (approved):

theo-mesnil:
😍

PR #883, opened 2021-08-12
feat: swiper to TS

guillaumewttj:


packages/Swiper/index.tsx
```diff
@@ -25,7 +68,6 @@ export const Swiper = forwardRef((props, ref) => {
     slidesToSwipe,
     ...rest
   } = props
-
   const translateX = -(pageIdx * 100)
 
   useEffect(() => {
```
guillaumewttj, line 73: On the lines below, should we just remove support for immutable objects since it makes no sense in the library ? And then let immutable users to use .toJS util in order to have a real javascript array ?

guillaumewttj: the link to the line (could not comment on unchanged parts 😅 ): https://github.com/WTTJ/welcome-ui/pull/883/files#diff-93657189125bc2e8fe169d4d98cbf890e29e16e1df9c9d2427fd455314cec5cdR75

packages/Swiper/index.tsx
```diff
@@ -1,12 +1,55 @@
...
+  setNumberOfSlides: (slidesLength: number) => void
+  setPageIdx: (idx: number) => void
+  slidesToShow: number
+  slidesToSwipe: number
+  nextButton?: ReactNode
+  prevButton?: ReactNode
+}
+
+export interface SwiperOptions {
+  renderPaginationItem?: (props: RenderPaginationProps) => ReactNode
```
theo-mesnil, line 43: We need to be same everywhere on Types import, React.ReactNode or ReactNode ?

theo-mesnil (approved):

packages/Swiper/index.tsx
mleralec: React.ReactNode is better I guess

guillaumewttj: done !

mleralec (commented):
Il manque le tsconfig.json :smile:

packages/Swiper/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Customizable design system with react • styled-components • styled-system and reakit.",
   "main": "dist/swiper.cjs.js",
   "module": "dist/swiper.es.js",
+  "types": "dist/index.d.ts",
```
mleralec, line 7:
```
"types": "dist/types/index.d.ts",
```

il manque également le script "types"

PR #886, opened 2021-08-12
feat: toast to TS

guillaumewttj:


theo-mesnil (approved):

packages/Toast/styles.ts
```diff
@@ -5,14 +5,16 @@ import { system } from '@welcome-ui/system'
 import { getVariantColor } from '@welcome-ui/utils'
 import { Alert } from '@welcome-ui/alert'
 
-export const Toast = styled(Box)(
+import { Variant } from '.'
+
+export const Toast = styled(Box)<{ isBottom: boolean }>(
```
theo-mesnil, line 10: is required ?

theo-mesnil (approved):

packages/Toast/styles.ts
guillaumewttj: Yes because the value passed as isBottom prop comes from this line: https://github.com/WTTJ/welcome-ui/pull/886/files#diff-31467f2d39297d5f1f992951613a8dcd5f9dd26ec52b04ec741ea8f0b6e3a163R34
So the value is always set as true / false

theo-mesnil: 👍

PR #919, opened 2021-09-13
feat: build without ts plugin & add build:types script

mleralec:

PR #920, opened 2021-09-14
chore: upgrade prettier to support typescript 4.4

guillaumewttj:
Waiting on #919


mleralec (approved):

Calvein:
Is it possible to update it in https://github.com/WTTJ/wttj-config-front and then update wttj-config-front here or we gonna have too many complications?

PR #951, opened 2021-11-03
feat: full screen modal on mobile

jeromeraffin:


packages/Modal/styles.js
```diff
@@ -56,6 +56,11 @@ export const Dialog = styled(ReakitDialog)(
       opacity: 1;
       margin-top: 0;
     }
+
+    @media (min-width: ${th(theme.screens.md)}px) {
```
Calvein, line 60: I was going to tell you that you could use `up` from xstyled but it's deprecated 😅 
https://xstyled.dev/docs/magic-styled-components/#up

Calvein (approved):

packages/Modal/styles.js
Calvein: But you can now do this now apparently:
```suggestion
    @media (min-width: md) {
```

https://xstyled.dev/docs/responsive-design/#using-media-queries-in-styled

cnairiwttj (approved):

packages/Modal/styles.js
jeromeraffin: Yep I tried, in this case I need to update xstyled because it's only available in the 2.3.0 > https://github.com/gregberge/xstyled/releases/tag/v2.3.0

jeromeraffin: as seen with @Calvein I updated with `up` and we'll migrate to the new writing in the next version

theo-mesnil (approved):

PR #962, opened 2021-11-18
feat: search ts

mleralec:


packages/Shape/styles.ts
```diff
@@ -10,7 +10,7 @@ const shapeStyles = (w: string, h: string, shape: ShapeType) => css`
   border-radius: ${shape === 'circle' && '50%'};
 `
 
-export const Shape = styled(Box)<{ shape: ShapeType }>(
+export const Shape = styled(Box)<{ shape?: ShapeType }>(
```
guillaumewttj, line 13: does shape prop need to be optional here ? since there is a default value on the prop forwarded as value here in the Shape/index.ts

guillaumewttj (approved):

theo-mesnil (approved):

PR #966, opened 2021-11-22
fix: persist Themes subfolders

guillaumewttj:
Before / after on the circle ci machine:
Capture d’écran 2021-11-22 à 09 12 06


mleralec (approved):
:fire:

theo-mesnil (approved):
🔥

PR #969, opened 2021-11-23
feat: custom forwardRef to inject AsComponent props

mleralec:
**OLD**
```
<Button as={props => <Link href="..." target="_blank" />} variant="primary" />
```

**NOW**
```
// error -> href/target does not exist on ButtonProps
<Button href="..." target"_blank" variant="primary" />

// good -> LinkProps are injected on ButtonProps
<Button as={Link} target="_blank" variant="primary" />
```


theo-mesnil (approved):
So coooool 😍

cnairiwttj (approved):

PR #972, opened 2021-11-25
fix: v4.0.0-alpha.8

mleralec:


Calvein (approved):

packages/System/index.tsx
```diff
@@ -149,3 +156,13 @@ export const forwardRef = <Component extends As, Props = {}>(
 ): CreateWuiComponent<Component, Props> => {
   return React.forwardRef(component)
 }
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const forwardFieldRef = <Component extends As, Props = {}>(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  component: React.ForwardRefRenderFunction<any, Props>
```
cnairiwttj, line 163: What about unknown instead of any?

cnairiwttj (approved):

PR #994, opened 2022-01-19
fix: tooltip v4 (rm wrapping div + types)

cnairiwttj:


packages/Tooltip/index.tsx
```diff
@@ -133,7 +133,7 @@ export type PlacementOptions =
   | 'left-start'
 
 export interface TooltipOptions {
-  children: React.ReactElement
+  children: React.ReactElement | React.ReactNode
```
Calvein, line 136: I'm pretty sure that `ReactElement` is in `ReactNode`
```suggestion
  children: React.ReactNode
```

Calvein (approved):

mleralec (approved):

packages/Tooltip/index.tsx
cnairiwttj: You're absolutely right sire, thanks @Calvein

PR #995, opened 2022-01-19
Fix slow toggle

Calvein:
Toggle acts weirdly when there is a performance bottleneck.

In this case, it's caused by algolia rerendering wayyy too many things:

[before](https://user-images.githubusercontent.com/243189/150114372-06217b86-c079-47c8-8261-92be76d5c923.mov)

I first used `:checked` & `:disabled` pseudo-classes to prevent having css generated at run time but it wasn't enough, the main culprit being the transition on the `left` property.
Here's a fixed version (still slow but not wonky):

[after](https://user-images.githubusercontent.com/243189/150114377-3507d52b-14e9-4a9b-9eec-9b372c33794e.mov)


mleralec (approved):
:rocket:

P1X3L (approved):

PR #996, opened 2022-01-19
fix:Add ttl on welcome-ui cdn font deploy

maxd-wttj:
Hi,
we currently push fonts to s3 without setting a cache-control max-age parameter. This behaviour is suboptimal and should be changed.


Calvein (approved):
👌

.circleci/config.yml
```diff
@@ -72,7 +72,8 @@ jobs:
           from: packages/IconFont/fonts
           to: s3://welcome-ui/public/fonts/$BRANCH/$ICON_FONT_HASH/
           arguments: |
-            --metadata GitCommit=$CIRCLE_SHA1 --delete
+            --metadata GitCommit=$CIRCLE_SHA1 --delete \
+            --cache-control "max-age=2678400"
```
guillaumewttj, line 76: I think we also have a script `webfont:deploy` that also uses aws cli to upload font files. 
Since it is already handled by Ci, maybe we can just remove webfont:deploy from `release` script

guillaumewttj (changes requested):

.circleci/config.yml
maxd-wttj: are you referring to this script: https://github.com/WTTJ/welcome-ui/blob/master/scripts/webfont-deploy.sh ?
I don't get where it is used exactly. Maybe @anucreative can tell us more about it since he wrote it.

guillaumewttj: yes it's this one. when i talk about scripts, its about npm scripts located here: https://github.com/WTTJ/welcome-ui/blob/master/package.json#L7

P1X3L (approved):

.circleci/config.yml
guillaumewttj: Summary of our meeting w/ @maxd-wttj, we should remove the call to `webfont:deploy` when executing a `release` or `release:prerelease` to let the ownership to the CI on how to deploy the webfont w/ wich options.
@anucreative is that ok for you ?

anucreative: Yeah, that makes sense. The idea is that CI takes care of releases when we merge into master ?

guillaumewttj: yes, thats it !

maxd-wttj: Hi guys!
I also think it's best to keep the webfont deploy outside of your release script. Should we merge this PR now? It will improve loading time on pages that use this asset.

guillaumewttj: I'm just adding a commit to this PR to remove the duplication !

guillaumewttj: finally created this PR to have separated reviews: https://github.com/WTTJ/welcome-ui/pull/1001

anucreative: On peut le merger maintenant n'est-ce pas @guillaumewttj

anucreative, line 76: Que penses-tu @maxd-wttj de rajouter `immutable` donné qu'il y a un hash ?
```suggestion
            --cache-control "public, max-age=2678400, immutable"
```

anucreative (changes requested):

.circleci/config.yml
maxd-wttj: C'est une bonne idée effectivement, je pense qu'on peut faire comme ça

mleralec (approved):

anucreative (approved):

guillaumewttj (approved):

PR #1104, opened 2022-03-29
fix: add async/await inline polyfill

mleralec:
Currently, if you build the `@welcome-ui/search` package (the only one that uses **async/await** in our components), in the output, we can find this : 

```js
  var searchResults = useCallback( /*#__PURE__*/function () {
    var _ref2 = _asyncToGenerator( /*#__PURE__*/regeneratorRuntime.mark(function _callee(value) {
      var data;
      return regeneratorRuntime.wrap(function _callee$(_context) {
        while (1) {
        ...
```

But there is no import of `regeneratorRuntime` because we have babelHelpers in `inline` mode. 💀 :boom:

This new polyfill will create `_async` and `_await` function in the current file and use them like this :
```js
function _await(value, then, direct) { ... }
function _async(f) { ... }

 var searchResults = useCallback(_async(function (value) {
    return _invokeIgnored(function () {
      if (minChars === 0 || (value === null || value === void 0 ? void 0 : value.length) >= minChars) {
        return _await(search(value), function (data) {
        ...
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,7 @@
     "aws-sdk": "^2.1094.0",
     "babel-jest": "27.5.1",
     "babel-plugin-annotate-pure-calls": "^0.4.0",
+    "babel-plugin-transform-async-to-promises": "^0.8.18",
     "colors": "^1.4.0",
     "conventional-github-releaser": "^3.1.3",
     "date-fns": "^2.28.0",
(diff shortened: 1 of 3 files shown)


cnairiwttj (approved):

mleralec:
We should not include **babel-plugin-transform-async-to-promises**, it's related to @rollup/plugin-babel, check https://github.com/rollup/plugins/issues/1155

mleralec:
We should not include **babel-plugin-transform-async-to-promises**, it's related to @rollup/plugin-babel, check https://github.com/rollup/plugins/issues/1155

theo-mesnil (approved):

PR #1261, opened 2022-05-05
refactor: system

mleralec:
We need to use `system` provided from `@xstyled/system` to maintain consistency

`@welcome-ui/system` no longer exports `system/componentSystem/wrapperSystem`

It's old code that needs to be cleaned up

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Here you'll find all the core components you need to create a delightful webapp.
 1 - Install the **peer dependencies** listed below:
 
 ```bash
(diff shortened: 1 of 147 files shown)


cnairiwttj (approved):

guillaumewttj (commented):
it's a breaking change, so do we want to put it in a minor version?

mleralec:
> it's a breaking change, so do we want to put it in a minor version?

@guillaumewttj Yes 👍🏻

PR #1290, opened 2022-05-16
feat: allow to use xstyled states

mleralec:
```jsx
<Box bg={{ _: 'primary.500', hover: 'dark.500' }}>Hover me</Box>
```

This will work now

cf https://xstyled.dev/docs/hover-focus-other-states


theo-mesnil (approved):
😍

PR #1317, opened 2022-05-24
feat: handle disabled options in select v.4

RobelTekle:
- add disabled in Option
- mimic `isMultiple` style and behavior (when user try to select an already selected option)

### DIFF
--- a/packages/Select/index.test.tsx
+++ b/packages/Select/index.test.tsx
@@ -197,7 +197,7 @@ test('<Select isMultiple> can select multiple items', () => {
   expect(tags.map(tag => tag.textContent)).toStrictEqual(['February', 'March', 'April'])
 })
 
-test('<Select isMultiple> cannot selet disabled items', () => {
+test('<Select isMultiple> cannot select disabled items', () => {
   const { getAllByRole, getByRole, getByTestId } = render(
     <Select
       dataTestId="select"
--- a/packages/Select/index.test.tsx
+++ b/packages/Select/index.test.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { fireEvent } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import capitalize from 'lodash.capitalize'
@@ -7,7 +7,17 @@ import { DateIcon } from '@welcome-ui/icons.date'
 
 import { render } from '../../utils/tests'
 
-import { Option, Select } from './index'
+import { Option, Select, SelectProps } from './index'
+
+const SelectWrapper: React.FC<SelectProps> = props => {
+  const [value, setValue] = useState<SelectProps['value']>()
+
+  const handleChange: SelectProps['onChange'] = newValue => {
+    setValue(newValue as SelectProps['value'])
+  }
+
+  return <Select onChange={handleChange} value={value} {...props} />
+}
 
 const MONTHS = [
   'january',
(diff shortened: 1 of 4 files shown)


packages/Select/index.test.tsx
```diff
@@ -7,7 +7,17 @@ import { DateIcon } from '@welcome-ui/icons.date'
...
 import { render } from '../../utils/tests'
 
-import { Option, Select } from './index'
+import { Option, Select, SelectProps } from './index'
+
+const SelectWrapper: React.FC<SelectProps> = props => {
+  const [value, setValue] = useState<SelectProps['value']>()
+
+  const handleChange: SelectProps['onChange'] = newValue => {
+    setValue(newValue as SelectProps['value'])
```
cnairiwttj, line 16: isn't newValue already typed with `SelectProps['onChange']` ?

packages/Select/index.test.tsx
```diff
@@ -173,6 +197,32 @@ test('<Select isMultiple> can select multiple items', () => {
   expect(tags.map(tag => tag.textContent)).toStrictEqual(['February', 'March', 'April'])
 })
 
+test('<Select isMultiple> cannot selet disabled items', () => {
```
cnairiwttj, line 200: Just a little typo here: `cannot select` but it's really a detail!

packages/Select/index.tsx
```diff
@@ -329,6 +334,7 @@ export const Select = forwardRef<'input', SelectProps>(
                                   const isItemSelected = isValueSelected(option.value, selected)
                                   return (
                                     <S.Item
+                                      $disabled={option.disabled}
```
cnairiwttj, line 337: Could you please tell me why there is a $ in front of disabled?

packages/Select/index.tsx
```diff
@@ -349,6 +355,7 @@ export const Select = forwardRef<'input', SelectProps>(
                           const isItemSelected = isValueSelected(result.value, selected)
                           acc.itemsToRender.push(
                             <S.Item
+                              $disabled={result.disabled}
```
cnairiwttj, line 358: same question here for the $

packages/Select/styles.ts
```diff
@@ -91,11 +92,13 @@ export const Item = styled.li(
     isHighlighted: boolean
     isMultiple: boolean
     isSelected: boolean
+    $disabled?: boolean
```
cnairiwttj, line 95: I don't think the $ is necessary here?

packages/Select/styles.ts
```diff
@@ -91,11 +92,13 @@ export const Item = styled.li(
...
   }) => css`
     color: nude.800;
     ${isHighlighted && th('defaultFields.select.highlighted')};
     ${isSelected && !isMultiple && th('defaultFields.select.selected')};
     ${isSelected && isMultiple && !allowUnselectFromList && th('defaultFields.select.existing')};
+    ${$disabled && th('defaultFields.select.disabled')};
```
cnairiwttj, line 101: +1 for the $

packages/Select/index.tsx
theo-mesnil: normally is to not be on the dom, only for the styled component

RobelTekle: it's to prevent `disabled` to be passed to the <li> in the DOM. Otherwise it will have a different behavior.

cnairiwttj (approved):

theo-mesnil (approved):

PR #1425, opened 2022-07-13
fix: allow styled props on dropdown menu

guillaumewttj:


packages/DropdownMenu/styles.ts
```diff
@@ -15,4 +15,5 @@ export const Inner = styled(Box)`
   &[hidden] {
     display: none;
   }
+  ${system}
```
theo-mesnil, line 18: add line break

theo-mesnil (approved):

PR #1493, opened 2022-08-16
fix: hydration initial UI

Geminii:
> **Note**
> Interface hydration problems come from the fact that NextJs will make a `diff` between what was generated on the server side and what is generated on the client side.

What is still problematic : 
- [x] The `Drawer.Backdrop` of [Header](https://github.com/WTTJ/welcome-ui/blob/master/docs/components/Header/index.js#L56)
- [x] The `Tooltip` present into the properties table (when the property is required)

### DIFF
--- a/docs/components/Colors.js
+++ b/docs/components/Colors.js
@@ -8,11 +8,17 @@ import { Text } from '@welcome-ui/text'
 function Wrapper({ name, theme, colorsObject }) {
   const colors = colorsObject || theme.colors[name]
   const isLight = name === 'light'
+  const textColor = isLight ? 'light.900' : 'dark.300'
 
   return (
     <Box>
       <Card mt="xl">
-        <Card.Body as={Grid} templateColumns="repeat(auto-fit, minmax(250px, 1fr))" gap="xl" backgroundColor={isLight && 'dark.900'}>
+        <Card.Body
+          as={Grid}
+          templateColumns="repeat(auto-fit, minmax(250px, 1fr))"
+          gap="xl"
+          backgroundColor={isLight && 'dark.900'}
+        >
           {Object.entries(colors).map(([key, depth]) => (
             <Box display="flex" alignItems="center" key={`${name}.${key}`}>
               <Box
@@ -24,15 +30,10 @@ function Wrapper({ name, theme, colorsObject }) {
                 w={60}
               />
               <Box px="sm">
-                <Text
-                  variant="h6"
-                  as="span"
-                  textTransform="capitalize"
-                  color={isLight && 'light.900'}
-                >
+                <Text variant="h6" as="span" textTransform="capitalize" color={textColor}>
                   {colorsObject ? key : `${name} ${key}`}
                 </Text>
-                <Text color={isLight ? "light.900" : "dark.300"} as="span" fontSize="sm" mt={3}>
+                <Text color={textColor} as="span" fontSize="sm" mt={3}>
                   {depth}
                 </Text>
               </Box>
--- a/docs/components/Colors.js
+++ b/docs/components/Colors.js
@@ -8,7 +8,7 @@ import { Text } from '@welcome-ui/text'
 function Wrapper({ name, theme, colorsObject }) {
   const colors = colorsObject || theme.colors[name]
   const isLight = name === 'light'
-  const textColor = isLight ? 'light.900' : 'dark.300'
+  const textColor = { ...(isLight && { color: 'light.900' }) }
 
   return (
     <Box>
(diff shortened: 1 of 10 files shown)


docs/components/Mdx/Props.js
```diff
@@ -120,11 +120,9 @@ export function Props({ propTypes }) {
                     </Table.Td>
                     <Table.Td textAlign="center" w="15%">
                       {required && (
-                        <Tooltip content="is required">
```
theo-mesnil, line 123: why did you removed tooltip ?

docs/components/Colors.js
```diff
@@ -24,15 +30,10 @@ function Wrapper({ name, theme, colorsObject }) {
                 w={60}
               />
               <Box px="sm">
-                <Text
-                  variant="h6"
-                  as="span"
-                  textTransform="capitalize"
-                  color={isLight && 'light.900'}
-                >
+                <Text variant="h6" as="span" textTransform="capitalize" color={textColor}>
```
theo-mesnil, line 33: is not the same rule than before

docs/components/Mdx/Props.js
theo-mesnil: we've got a bug in this component i think, try to add a div between tooltip and tag

Geminii: no improvement there : 

> Warning: An error occurred during hydration. The server HTML was replaced with client content in "div"`. 

> See more info here: https://nextjs.org/docs/messages/react-hydration-error

> Warning: Received 'true' for a non-boolean attribute 'modal'.
If you want to write it to the DOM, pass a string instead: modal="true" or modal={value.toString()}.`

Is this a problem related to Reakit ?

docs/components/Colors.js
Geminii: In the case of dark mode, we don't pass color for the `Text` component and got an error because we pass "undefined". What we need to use in this another case ?

docs/components/Mdx/Props.js
cnairiwttj: Hey @Geminii I think I've already had this issue when the first child of `tooltip`is a div. Could you please try `<Tooltip content="is required" fixed>` to avoid this error?

docs/pages/upgrades/v4.mdx
```diff
@@ -0,0 +1,106 @@
...
+### Text component
+
+We reduce the size (and variant) of headings on mobile :
+
+- h0 > h1
+- h1 > h2
+- h2 > h3
+- h3 > h4
+- h4 > h5
+- h5 > h6
```
cnairiwttj, line 69: Not sure what this means ? h0 is replaced with h1 size, and so on?

docs/components/Colors.js
cnairiwttj: @Geminii `{...(isLight  && { color: 'light.900' })}` could work?

cnairiwttj (commented):
Hey @Geminii 
Not so much things to add after Theo's review.
Could you please next time make a smaller PR please? It was difficult to review (made my browser crashed several times...)
Thanks 🔥

Geminii:
> Hey @Geminii Not so much things to add after Theo's review. Could you please next time make a smaller PR please? It was difficult to review (made my browser crashed several times...) Thanks 🔥

Hey @cnairiwttj 
Sorry for this, but there was few changes when I opened my PR (need to rebase from `v5` 😇)

docs/components/Mdx/Props.js
Geminii: @cnairiwttj We continued to test and investigate with @theo-mesnil but no result for the moment. I tried to add `fixed` like you suggest but no improvement 😭

docs/pages/upgrades/v4.mdx
Geminii: Don't have precision about this modification (not related to this PR). It was a previous changes due to spacing modification.

docs/components/Colors.js
Geminii: This works, thanks @cnairiwttj 🙏

docs/components/Mdx/Props.js
theo-mesnil: This information is useless on the ux we removed it for now :)

theo-mesnil (approved):

cnairiwttj (approved):

docs/components/Mdx/index.js
```diff
@@ -31,12 +31,12 @@ export const MDXComponents = {
   p: props => <Text as="p" lineHeight="1.5" pt="lg" {...props} />,
   strong: props => <Box as="strong" fontWeight="bold" {...props} />,
   ul: props => (
-    <Box as="p" mb="sm">
+    <Box as="div" mb="sm">
```
guillaumewttj, line 34: Useless as div

theo-mesnil (approved):

PR #1503, opened 2022-08-16
feat: new visual for tag component

### LINKED ISSUE #1471: Tag
[Link to Figma](https://www.figma.com/file/jAg2mzxXFZocK1MZOmX6IA/Data-display?node-id=3%3A22)

Geminii:
fixes #1471 

- [x] New visual (colors / spacing)
- [x] Transition between default and hover state
- [x] Icon size adjustment
- [x] Upgrade documentation

https://feat-tag-component--welcome-ui.netlify.com


packages/Core/theme/tags.ts
```diff
@@ -107,26 +101,58 @@ export const getTags = (theme: WuiTheme): ThemeTags => {
...
+      info: {
+        borderColor: colors.info[500],
+      },
+      1: {},
+      2: {},
+      3: {},
+      4: {},
+      5: {},
+      6: {},
+      7: {},
```
theo-mesnil, line 130: i think we need also hover on this variants

packages/Core/theme/tags.ts
```diff
@@ -107,26 +101,58 @@ export const getTags = (theme: WuiTheme): ThemeTags => {
...
+    },
     sizes: {
       xs: {
-        padding: `0 ${space.xxs}`,
+        padding: `${space.xxs} ${space.xs}`,
         height: sizes.xs,
-        fontSize: toRem(10),
+        fontSize: fontSizes.xs,
+        gap: toRem(4),
```
theo-mesnil, line 137: we need to get the value from spaces.xs

packages/Tag/styles.ts
```diff
@@ -71,6 +78,10 @@ export const Tag = styled.div<StyledTagProps & WuiProps>(
         }
       }
     }
+
+    &:hover {
```
theo-mesnil, line 82: we need to activate the hover style if is a link or got an onClick property ?

theo-mesnil: and add a cursor pointer?

Geminii: We already got a pointer for `onRemove` action but rest of the time, we don't need it because it's not a link no ?

packages/Core/theme/tags.ts
Geminii: For primary and subs, we use [withoutVisibleBorder()](https://github.com/WTTJ/welcome-ui/pull/1503/files#diff-8c6efd32613876e4549318477a7d44c7ab5e0de209b90e0add8c367514d0b4fcR65). Do we need to add `borderColor` on hover ?

packages/Tag/styles.ts
theo-mesnil: It can be a link (for example list of tags to go on a page) or can have an onClick to, we do not want the hover effect if is not clickable

packages/Core/theme/tags.ts
theo-mesnil: We i think a dark.200 for example to get the same things on subs like the others on hover

docs/pages/components/tag.mdx
```diff
@@ -63,46 +62,50 @@ Use size property with `xs`, `sm`, `md` or `lg`.
...
-  Long text
+<Tag size="md">
+  1
 </Tag>
 ```
 
-Set `shape` prop to `circle` to set a border-radius `50%`.
+## OnClick
+
+Simply add a function `OnClick` to catch the action with an hover effect.
```
cnairiwttj, line 101: Hey, here it's an `onClick` function. It's not very different but I think it could be confusing

docs/pages/upgrade.mdx
```diff
@@ -14,6 +14,13 @@ Version 5: Rework of our foundations
 - The `quaternary` variant is now named `ghost`.
 - The icon size now adjusts to the size of the button.
 
+#### Tag
+
+- We have adjust colors and spacing.
```
cnairiwttj, line 19: We have adjust**ed** (again, not a important fix sorry)

packages/Tag/styles.ts
```diff
@@ -60,6 +76,12 @@ export const Tag = styled.div<StyledTagProps & WuiProps>(
       margin-right: xxs;
     }
 
+    & > svg:only-child,
+    & > svg:not(:only-child) {
```
cnairiwttj, line 80: I think I'm wrong but how is this rule different from `& > svg` ? (since it's targeting only-child AND not only-child?)

docs/pages/components/tag.mdx
Geminii: Totally agree with you 👍

docs/pages/upgrade.mdx
Geminii: Well seen 😇 🙏

packages/Tag/styles.ts
Geminii: You are not wrong, we could simplify it by `& > svg` 😅 quite simply 🙈

packages/Tag/styles.ts
Geminii: Examples has been added in the `onClick` section 😀

theo-mesnil (approved):

cnairiwttj (approved):

theo-mesnil (approved):

Issue #1472, opened 2022-08-02
Tabs


Jademhm:
[Link to Figma](https://www.figma.com/file/JrPBB5pItDO43GhHE7pT3K/Navigation?node-id=2%3A3)

theo-mesnil:
@Jademhm tu peux testée ici https://feat-tabs--welcome-ui.netlify.app/

Jademhm:
Here's the feedback:
* **Font size**
    * Medium → Body M = 16px
    * Small → Body S = 14px (≠ 16px currently)
* **Underline**
	* Not supposed to be 3px but 2px
* **Paddings**
    * Font from bottom → 12px for both sizes
    * Currently there’s 16px on Medium and 8px on Small
* **Colour**
    * Disabled → I don’t know what the colour is in the inspect, it returns me "rgba(0, 0, 0, 0.55)" which would be Dark 500 in our design records but it’s supposed to be Light 700 (#B3B3B3)

Vico4:
>   * Disabled → I don’t know what the colour is in the inspect, it returns me "rgba(0, 0, 0, 0.55)" which would be Dark 500 in our design records but it’s supposed to be Light 700 (#B3B3B3)

@Jademhm Light 700 works only on dark theme (it's white with a transparency so it appears white on the doc), is there a light theme color that fits ? Dark 500 was the closest visually. 
For everything else I do it today, thank you :)

Jademhm:
Default = Dark 700
Disabled = Dark 400

PR #1504, opened 2022-08-17
feat: tabs

Vico4:
Issue #1472 : 

- [x] size options 
- [x] spacing
- [x] color
- [x] icon/img/badge size and spacing

---
  

 👀 [Visit Preview](https://feat-tabs--welcome-ui.netlify.app)

### DIFF
--- a/packages/Core/theme/tabs.ts
+++ b/packages/Core/theme/tabs.ts
@@ -2,6 +2,8 @@ import { CSSObject } from '@xstyled/styled-components'
 
 import { WuiTheme } from './types'
 
+type Size = 'sm' | 'md'
+
 export type ThemeTabs = {
   tabsBorder: {
     horizontal: CSSObject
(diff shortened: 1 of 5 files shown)


packages/Core/theme/tabs.ts
```diff
@@ -75,5 +81,22 @@ export const getTabs = (theme: WuiTheme): ThemeTabs => {
...
     },
+    size: {
+      sm: {
+        paddingBottom: toRem(9),
+        marginRight: toRem(12),
+      },
+      md: {
+        paddingBottom: toRem(15),
+        marginRight: toRem(24),
+      },
```
theo-mesnil, line 92: Tou need to get this values from theme

packages/Tabs/styles.ts
```diff
@@ -26,12 +33,11 @@ export const Tab = styled.button`
   display: flex;
   align-items: center;
   flex: none;
-  margin-right: xl;
   padding: lg 0;
   transition: medium;
   text-transform: none;
   cursor: pointer;
-  ${system};
+  gap: 4;
```
theo-mesnil, line 40: Get this value from theme

packages/Core/theme/tabs.ts
Vico4: when value doesn't exist in theme (ie. 9, 15), I go for the closest one or I keep toRem() with requested value ?

theo-mesnil: Closest, i think the component is not fully good on design team

theo-mesnil (approved):

PR #1554, opened 2022-08-29
feat: tooltip

Isaacmaamouche:
- Updated the styling
- Added the fade/slide-in animation with 4 different directions depending on the placement property
- Slide-in disabled in the property `fixed` is used or if placement's value is  `auto-*`

👀 [Visit Preview](https://v5-tooltip-1465--welcome-ui.netlify.app)

### DIFF
--- a/docs/pages/components/tooltip.mdx
+++ b/docs/pages/components/tooltip.mdx
@@ -48,38 +48,108 @@ Set a `max-width` for long text
 
 ## With a fixed position
 
-Using the `fixed` prop in combination with `placement` will enable a translate animation on your tooltip. The direction works with these values for `placement` : `top-*`, `right-*`, `bottom-*` and `left-*`.
+Using the `fixed` prop in combination with `placement` will enable a translate animation on your tooltip. The direction works with these values for `placement`.
 
 ```jsx
-<Flex gap="lg">
-  <Tooltip fixed placement="top" content="🔥🔥🔥🔥🔥🔥🔥">
+<Flex gap="lg" flexWrap='wrap' justifyContent='center'  >
+
+  <Tooltip fixed placement="top-start" content="🔥🔥🔥🔥🔥🔥🔥">
     <Button>
-      <span>Coming from the top</span>
-      <ArrowUpIcon />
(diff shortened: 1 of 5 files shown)


packages/Tooltip/styles.ts
```diff
@@ -2,9 +2,43 @@ import styled, { css, system, th } from '@xstyled/styled-components'
...
+}
+type TransformDirectionKey = keyof typeof transformDirection
+
+export const FadeIn = styled.div<{
+  placement?: PlacementOptions
+  fixed?: TooltipOptions['fixed']
+}>`
+  ${th('tooltips')};
+  ${system};
+  background-color: rgba(33, 33, 33, 0.9);
```
Vico4, line 25: get from theme if possible

Vico4 (changes requested):
on the demo there is an extra line on every tooltip

packages/Tooltip/styles.ts
Leiksa: background-color is given on the Tooltip figma -> https://www.figma.com/file/TC5aahnPVDU1hhBcff037e/Feedback?node-id=2%3A890

Leiksa (changes requested):
Screenshot 2022-08-30 at 12 05 32

You seems to have a padding problem or something else. (cf screen)

Vico4 (approved):
🔥

Leiksa (approved):

theo-mesnil (approved):

docs/pages/components/tooltip.mdx
```diff
@@ -46,6 +46,109 @@ Set a `max-width` for long text
 </Tooltip>
 ```
 
+## With a fixed position
+
+Using the `fixed` prop in combination with `placement` will enable a translate animation on your tooltip. The direction works with these values for `placement`.
+
+```jsx
+<Flex gap="lg" flexWrap='wrap'>
+
```
theo-mesnil, line 55: Remove all useless break line and align code (like on vscode ;))

theo-mesnil (approved):

PR #1555, opened 2022-08-30
feat: helper for theme configuration

Geminii:
### Problem : 

It's a bit difficult to remember the value and convert rem value to pixel regarding spacing, screens, font sizes, font weight, etc. it's like constant mental gymnastics 🤸‍♀️ 

### Solution purposed : 

Create an helper to display converted values 🚀 and be relaxed now 😎
How to display the helper : `CMD + I`

https://user-images.githubusercontent.com/9429420/187430395-9ba4e2ed-a8c6-4e6b-91c4-5e13c1cc0305.mov

### DIFF
--- a/docs/components/Layouts/Docs/index.js
+++ b/docs/components/Layouts/Docs/index.js
@@ -3,6 +3,7 @@
 import React from 'react'
 import { Box } from '@welcome-ui/box'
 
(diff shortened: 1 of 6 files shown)


mleralec (approved):
👏🏻

PR #1558, opened 2022-09-07
feat: new slider component

Leiksa:


packages/Slider/styles.ts
```diff
@@ -0,0 +1,192 @@
+import styled, { css, th } from '@xstyled/styled-components'
+import { Box } from '@welcome-ui/box'
+import { Hint } from '@welcome-ui/hint'
+
+type BorderProps = {
+  borderSelectorColor?: string
+}
+
+export const Slider = styled.input<BorderProps>(
+  ({ borderSelectorColor, disabled }) => css`
```
cnairiwttj, line 10: Shouldn't we giving a default value for borderSelectorColor as it's not mandatory?

packages/Slider/styles.ts
```diff
@@ -0,0 +1,192 @@
...
+        &::-moz-range-thumb {
+          ${th('sliders.focused.disabled')};
+        }
+      }
+    `};
+  `
+)
+
+export const RangeInput = styled.input<BorderProps>(
+  ({ borderSelectorColor, disabled }) => css`
```
cnairiwttj, line 66: Same question here as above 🔝

packages/Slider/index.test.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import { Slider } from './index'
+
+describe('<Slider>', () => {
+  it('should render correctly', () => {
+    const handleChange = jest.fn()
+    const { container } = render(<Slider max={100} min={0} onChange={handleChange} value={50} />)
+
+    expect(container)
+  })
```
cnairiwttj, line 13: Maybe you could add more tests for the Slider component? Check if it correctly renders range if you add it, if the value changes when arrows left or right are pressed for instance...?

packages/Slider/Range.tsx
```diff
@@ -0,0 +1,276 @@
...
+    const tooltipMinRef = useRef<HTMLOutputElement>(null)
+    const tooltipMaxRef = useRef<HTMLOutputElement>(null)
+    const [tooltipMinVisible, setTooltipMinVisible] = useState<boolean>(false)
+    const [tooltipMaxVisible, setTooltipMaxVisible] = useState<boolean>(false)
+
+    const handleMinValue = (e: ChangeEvent<HTMLInputElement>) => {
+      // Prevents the min value from being above the max
+      const value = round(Math.min(parseInt(e.target.value, 10), maxValue - 1), step)
+      setMinValue(value)
+      e.target.value = value.toString()
```
cnairiwttj, line 60: Could you please tell me what is the purpose of this line `e.target.value = value.toString()`

packages/Slider/Range.tsx
```diff
@@ -0,0 +1,276 @@
...
+    const maxValueRef = useRef<HTMLInputElement>(null)
+    const range = useRef<HTMLDivElement>(null)
+    const tooltipMinRef = useRef<HTMLOutputElement>(null)
+    const tooltipMaxRef = useRef<HTMLOutputElement>(null)
+    const [tooltipMinVisible, setTooltipMinVisible] = useState<boolean>(false)
+    const [tooltipMaxVisible, setTooltipMaxVisible] = useState<boolean>(false)
+
+    const handleMinValue = (e: ChangeEvent<HTMLInputElement>) => {
+      // Prevents the min value from being above the max
+      const value = round(Math.min(parseInt(e.target.value, 10), maxValue - 1), step)
```
cnairiwttj, line 58: No need to verify if `isNan(e.target.value)`? (Same question for handleMaxValue)

packages/Slider/Range.tsx
```diff
@@ -0,0 +1,276 @@
...
+        if (tooltipMinRef.current) {
+          const thumbWidth = 20
+          const fraction = getPercent(minValue) / 100
+          tooltipMinRef.current.style.left = `calc(${fraction * 100}% + ${
+            (0.5 - fraction) * thumbWidth
+          }px)`
+        }
+
+        if (tooltipMaxRef.current) {
+          const thumbWidth = 20
```
cnairiwttj, line 112: Maybe you could extract this const above the component? I don't know if there is another const like this one, which does not rely on any variable, but you can extract them from the component render.

packages/Slider/Range.tsx
Leiksa: As the min & max are forced to be `number`, I don't think we have to check if the `parseInt` result to a `NaN`.
We need to do this conversion because of the `input` doing the inverse thing (use `number` attributes & convert them to `string` for event `value`)

packages/Slider/Range.tsx
Leiksa: The line `58` handle the `step` logic & make sure the `minValue` doesn't go above the `maxValue`. If we don't update the event `value` & you try to catch the `event` or put a ref on the `Slider` you will have the value before the check

packages/Slider/index.test.tsx
Leiksa: Yes, I wanted to ask a front developer from another squad to help me doing some tests since I don't have a lot of experience doing it :D

packages/Slider/styles.ts
cnairi: The issue I see here is that if borderSelectorColor is undefined, you will have `border-color: undefined`.
It will cause an error in your console I think. 
If you set borderSelectorColor = transparent by default, it will not override the border-color provided when borderSelectorColor is undefined, and prevent any console error. 
As BorderProps' borderSelectorColor is not required, it's possible to receive undefined as a value I guess.

packages/Slider/index.test.tsx
cnairi: Hey @Leiksa, I can help you with your test if you want, maybe on Wednesday? What you can do by then is writing all the test scenarii you would like to experiment :) And we'll try to implement them together

theo-mesnil (approved):
😻

PR #1561, opened 2022-09-08
feat: colors autocomplete

mleralec:
Well 😈 

Currently, we don't have auto-completion in our JSX components on color props like `color`, `backgroundColor`, `borderColor`...

This is because **xstyled** expects to find strings inside `theme.colors`, we can see that in the [default theme of xstyled](https://github.com/gregberge/xstyled/blob/main/packages/system/src/defaultTheme.ts#L71).

In the example below, you can see the auto-completion working: 
([full example available here](https://codesandbox.io/s/autocompletion-colors-so88rn?file=/src/index.tsx))

https://user-images.githubusercontent.com/36373219/189138725-abb21228-8d42-4d9c-9dc2-0da98d2a62f5.mov

The solution is to replace every keys of `theme.colors` with a `-` instead of a `.` 
`primary.500` -> `primary-500`


Calvein (approved):
😱

docs/pages/theming/basics.mdx
```diff
@@ -132,10 +126,10 @@ We use [screens from xstyled](https://xstyled.dev/docs/responsive-design/). Defa
   alignItems="center"
   justifyContent="center"
   textAlign="center"
-  color={{ xs: 'light.900', lg: 'dark.900' }}
+  color={{ xs: 'light-900', lg: 'dark-900' }}
   fontSize={{ lg: 'h2' }}
   h={{ xs: '10rem', lg: '20rem' }}
-  backgroundColor={{ xs: 'sub.1', lg: 'primary.100' }}
+  backgroundColor={{ xs: 'sub.1', lg: 'primary-100' }}
```
cnairi, line 132: `sub.1` or `sub-1`?

mleralec: good catch 👀

cnairi (approved):

cnairi (approved):
🔥 (I hate you 🤣 )

theo-mesnil (approved):
💥

PR #1588, opened 2022-09-22
feat: rework badge for v5

theo-mesnil:
Signed-off-by: Théo Mesnil <mesniltheo@gmail.com>

### DIFF
--- a/packages/Badge/index.tsx
+++ b/packages/Badge/index.tsx
@@ -1,12 +1,62 @@
 import React from 'react'
-import { Tag, TagOptions } from '@welcome-ui/tag'
 import { CreateWuiProps, forwardRef } from '@welcome-ui/system'
 
-export type BadgeOptions = Omit<TagOptions, 'size'>
+import * as S from './styles'
+
+export type BadgeOptions = {
+  disabled?: boolean
+  shape?: 'circle' | 'square'
+  size?: 'sm' | 'md'
+  variant?: 'default' | 'primary'
+  // if a number is higher than 99, we replace this number by 99+
+  withNumberAbbreviation?: boolean
+  children: string | number
+}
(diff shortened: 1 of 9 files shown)


packages/Badge/index.tsx
```diff
@@ -1,12 +1,62 @@
...
+import * as S from './styles'
+
+export type BadgeOptions = {
+  disabled?: boolean
+  shape?: 'circle' | 'square'
+  size?: 'sm' | 'md'
+  variant?: 'default' | 'primary'
+  // if a number is higher than 99, we replace this number by 99+
+  withNumberAbbreviation?: boolean
+  children: string | number
```
Calvein, line 13: We don't want to allow any type of children?
e.g:

```jsx
<Badge>
  <FormattedMessage defaultMessage="New" id="new" />
</Badge>
```

If so, it should be:
```suggestion
  children: React.node
```

packages/Badge/index.tsx
```diff
@@ -1,12 +1,62 @@
...
+      size = 'md',
+      variant = 'default',
+      withNumberAbbreviation,
+      ...rest
+    },
+    ref
+  ) => {
+    let text
+    const isNumber = Number.isInteger(children)
+    const textLength = children.toString().length
```
Calvein, line 33: Ah that's why we want only strings or number

packages/Badge/index.tsx
theo-mesnil: yes, you can do with the intl hook instead

packages/Badge/index.tsx
theo-mesnil: yes :)

cnairiwttj (approved):

PR #1818, opened 2023-01-04
Fix: update forms

Vico4:


cnairiwttj (approved):

packages/Field/styles.ts
```diff
@@ -68,6 +68,14 @@ export const IconWrapper = styled.div<IconWrapperProps>(
     & > button {
       pointer-events: auto;
     }
+
+    ${size === 'xs' &&
+    css`
+      button {
+        width: 16;
+        height: 16;
```
theo-mesnil, line 76: Can you add this on field theme please ? :)

Vico4: added in button theme because it seemed to make more sense to change the size of the clear button and it had less impact on the types

theo-mesnil: Ok, you need to add some examples on the documentation, and d o not forget to add on all button theme this size (on iconOnly, on iconSize etc)

theo-mesnil (approved):

theo-mesnil (approved):

cnairiwttj (approved):

Issue #1922, opened 2023-02-08
Modal do not scroll on iOS


theo-mesnil:
On the Modal component, the modal do not scroll on iOS, it's a bug from ariakit.

PR #1923, opened 2023-02-08
fix: modal scroll on mobile device

Leiksa:
Related issue: #1922 

As Reakit doesn't handle scrolling content we have to forward the modalState to the Modal.Content in order to check when the modal is visible and enable the scroll.
_Reakit issue:_ https://github.com/ariakit/ariakit/issues/469

I preferred to use a context to keep the logic internally and not expose a misleading prop which could be use in the implementation and also as we render children and not the Modal.Content directly, thanks to the context we avoid using `children.map` and `cloneElement`.


theo-mesnil (approved):
🔥

Issue #1899, opened 2023-02-02
Swiper


jeromeraffin:
Update the swiper code to have more functionalities

PR #1932, opened 2023-02-10
feat: new swiper for wui v5

jeromeraffin:
Related issue: https://github.com/WTTJ/welcome-ui/issues/1899

### DIFF
--- a/docs/pages/components/swiper.mdx
+++ b/docs/pages/components/swiper.mdx
@@ -63,6 +63,56 @@ function() {
 }
 ```
 
+## First slide to show
+
+By providing `firstSlideToShow` with `2`, the swiper will begin on the second slide.
+
+```jsx
+function() {
+  return (
+    <Swiper h={400} firstSlideToShow={2} >
+      <img
+        style={{ maxWidth: '100%', maxHeight: '100%', objectFit: 'cover' }}
+        src="https://images.unsplash.com/photo-1564460549828-f0219a31bf90?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=1350&q=80"
+      />
+      <img
+        style={{ maxWidth: '100%', maxHeight: '100%', objectFit: 'cover' }}
+        src="https://images.unsplash.com/photo-1575489272413-cb506258027e?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=1350&q=80"
+      />
+      <img
+        style={{ maxWidth: '100%', maxHeight: '100%', objectFit: 'cover' }}
+        src="https://images.unsplash.com/photo-1541959833400-049d37f98ccd?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=1350&q=80"
+      />
+    </Swiper>
+  )
+}
+```
+
+## Centered slides
+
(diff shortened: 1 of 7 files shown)


docs/pages/components/swiper.mdx
```diff
@@ -19,238 +19,120 @@ import {
...
 
-## With custom pagination items
+## Without arrow buttons
 
-By providing the optional `renderPaginationItem` you can customise the defaults pagination bullets and render your own component. This callback takes a `props` param which is an object containning:
-
-- `idx`: the current pagination item
-- `pageIdx`: the currently selected slide (which you can use to show the active state for your pagination items)
-- `onClick`: callback to pass to your custom bullet
+By providing the optional `withArrows` with `[false, false]`, you can hide the navigation arrows on mobile/tablet and desktop.
```
cnairiwttj, line 118: I'm wondering for all these props if it wouldn't be clearer to have an object?
Like this:
```
withArrows = { mobile: false, desktop: false }
```
In my opinion this is more explicit but I don't know if it's more complex

packages/Swiper/src/index.tsx
```diff
@@ -1,175 +1,168 @@
...
-        >
-          {nextButton}
-        </S.Next>
-      )}
-    </S.Wrapper>
-  )
-})
+  const firstPageToShow = centeredSlides
+    ? Math.floor(slidesLength / currentSlidesPerView / 2)
+    : Math.ceil(firstSlideToShow / currentSlidesPerView) - 1
```
cnairiwttj, line 81: Could you please just explain to me what this calculation is doing?
I don't understand why in the first case we do not take into account the firstSlideToShow?

theo-mesnil: @jeromeraffin please add some description :)

docs/pages/components/swiper.mdx
jeromeraffin: Yep I agree, I changed all the props with an array for an object, thanks!

theo-mesnil (approved):
🔥

- Please add on upgrade.mdx page the breaking changed
- Add more description on complex function

packages/Swiper/src/index.tsx
```diff
@@ -1,175 +1,181 @@
...
-        >
-          {nextButton}
-        </S.Next>
-      )}
-    </S.Wrapper>
-  )
-})
+  const firstPageToShow = centeredSlides
+    ? // if centeredSlides is true, we calculate which number is the middle page
+      Math.ceil(numberOfPage / 2)
```
cnairiwttj, line 92: Thanks for the explanation, it's much clearer!
Maybe we could say it in the doc that the firstSlideToShow and centeredSlides props cannot be combined?
I don't know if you've already written it

jeromeraffin: Yes you're right, I just add the info on the doc that `firstSlideToShow` won't be used if centeredSlides is true

packages/Swiper/src/index.tsx
cnairiwttj: 🔥

docs/pages/components/swiper.mdx
cnairiwttj: 🔥  thank you!

packages/Swiper/src/index.tsx
cnairiwttj: That's perfect, thanks 🔥

cnairiwttj (approved):
🔥 Congrats for this huge work!

PR #1939, opened 2023-02-14
fix: modal header spacing

Leiksa:


packages/Modal/Content.tsx
```diff
@@ -51,7 +51,12 @@ export const Content = forwardRef<'div', ContentProps>(({ children, ...rest }, r
   const getStyles = (name?: string) => {
     if (name === 'Header') {
       return {
-        pb: components.includes('Body') || components.includes('Footer') ? space.lg : null,
+        // if the Modal have a Body but not a Footer || have a Footer but not a Body
```
cnairiwttj, line 54: Hey @Leiksa,

I've just checked on the current alpha with/without Body/Footer.
Is it really necessary to create different cases? Can't see the issue

Leiksa: Hello @cnairiwttj , 
We noticed a spacing problem on the modal header on wttj-front, indeed it was 16px instead of 32px.
Then, while checking the design system I noticed that the spacing was different if there was a footer or not (see the DS screenshot below). Available if you have any questions. 

Screenshot 2023-02-14 at 17 11 04
Screenshot 2023-02-14 at 17 10 57
adding was wrong

theo-mesnil (approved):

packages/Modal/Content.tsx
cnairiwttj: Ok thanks a lot for the explanation @Leiksa 
I honestly think it will be difficult to maintain but I understand the need. 
It would be nice to check with the design team how to simplify the design rule of the modal :)

cnairiwttj (approved):

packages/Modal/Content.tsx
Leiksa: completely agree with you!

PR #1960, opened 2023-02-22
chore: prepare to ariakit

### LINKED ISSUE #1913: Prepare to ariakit
We need to rework the way to manage the state from reakit librairy on our components to prepare the transition to ariakit (when it's no longer in alpha).

On the ariakit (reakit v2), the state is pass to component with the `state` property.

Old way:
```
const dialog = useDialogState();

<Modal {...dialog}>...</Modal>
```

New way:
```
const dialog = useDialogState();

<Modal state={dialog}>...</Modal>
```

Impacted components:

- [x] Accordion
- [x] Drawer
- [x] DropdownMenu
- [x] EmojiPicker
- [x] Modal (change modalState to state)
- [x] Popover
- [x] Tab

theo-mesnil:
Fixes #1913 

We need to prepare the migration to [ariakit](https://ariakit.org) (reakit v2), and this migration contains some breaking changes.
We don't want create a version 6 of WUI with the breaking changes when ariakit will be a release candidate (now is on alpha), so we take the lead and anticipate the breaking changes on the V5.

What we changed:
- now the state from a reakit state is not destructured on the component
```jsx
const tabState = useTabState()

<Tab.List state={tabState}>
  <Tab state={tabState}>
    ...
  </Tab>
</Tab.List>
<Tab.Panel state={tabState}>
  ...
</Tab.Panel>
```
- we deprecated the `visible` option, it's replace by the `open` option. _**It will be removed when we add ariakit on WUI**, so please changed now visible property to open if you use it :)_
```jsx
const modalState = useModalState({ open: true })

const { open } = modalState
```

Which components are impacted?
- Accordion
- Drawer
- DropdownMenu
- EmojiPicker
- Modal
- Popover
- Tab

### DIFF
--- a/docs/components/Header/ThemeSelector.js
+++ b/docs/components/Header/ThemeSelector.js
@@ -1,20 +1,20 @@
 import React from 'react'
 import { DropdownMenu, useDropdownMenuState } from '@welcome-ui/dropdown-menu'
 import { Box } from '@welcome-ui/box'
-import { StarIcon, SunIcon, CrescentMoonIcon, StarOutlineIcon } from '@welcome-ui/icons'
+import { CrescentMoonIcon, StarIcon, StarOutlineIcon, SunIcon } from '@welcome-ui/icons'
 import { Button } from '@welcome-ui/button'
 import { Badge } from '@welcome-ui/badge'
 
 import { useSetThemeContext, useThemeContext } from '../../context/theme'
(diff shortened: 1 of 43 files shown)


cnairiwttj (approved):

mleralec (approved):
:fire:

PR #1983, opened 2023-03-02
fix: swiper navigation with keyboard and possibility to change navigation size

jeromeraffin:


cnairiwttj (approved):
Hi Jay, I've just checked the previous of the swiper and I've realized that the default values for the object props are truncated.
Example:
`{ mobile: 1, tablet: 1, desk...`
Do you think it is possible to fix that?

PR #2010, opened 2023-03-16
chore: upgrade-packages-next script

mleralec:
Unfortunately, `yarn upgrade-interactive --latest` does not work with the prerelease versions...
It is very boring to upgrade welcome-ui v5 (still in alpha), this script should help 🤗 

After the script:
image

(don't forget to run a `yarn install` after running this script)


theo-mesnil (approved):
🔥

Issue #1912, opened 2023-02-07
Typescript error: condition on Modal.Content component


theo-mesnil:
On the Modal.Content component, sometime we need to add condition to show or hide a Modal.Footer for example. By doing this a typescript error pop: "No overload matches this call. Type 'boolean' is not assignable to type 'Element'."

```
<Modal.Content>
   {withFooter && <Modal.Footer>lorem</Modal.Footer>}
</Modal.Content>
```

Find an issue to avoid typescript error

PR #2017, opened 2023-03-21
fix: modal content children type

Leiksa:
Related issue: https://github.com/WTTJ/welcome-ui/issues/1912

### DIFF
--- a/packages/Modal/src/Content.tsx
+++ b/packages/Modal/src/Content.tsx
@@ -7,7 +7,7 @@ import * as S from './styles'
 import { useModal } from './context'
 
 export interface ContentOptions {
-  children: JSX.Element | JSX.Element[]
+  children: React.ReactNode
 }
 
 export type ContentProps = ContentOptions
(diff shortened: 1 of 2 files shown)


Vico4 (approved):

mleralec (commented):
we should add tests each time we fix an issue

RobelTekle (approved):

Leiksa:
> we should add tests each time we fix an issue

@mleralec I agree with you but here it's just a typescript error when rendering conditionally inside the Modal.Content, in fact the modal doesn't crash.

mleralec:
> > we should add tests each time we fix an issue
> 
> @mleralec I agree with you but here it's just a typescript error when rendering conditionally inside the Modal.Content, in fact the modal doesn't crash.

You can just add a test with conditional rendering in the `Modal.Content`, with this test, if one day we want to refactoring this part of the code, we will get a typescript error

theo-mesnil (approved):

mleralec (approved):

PR #2048, opened 2023-04-06
Feat: override modal.hide with onClose prop value

Isaacmaamouche:


packages/Modal/src/index.tsx
```diff
@@ -66,25 +66,26 @@ const ModalComponent = forwardRef<'div', ModalProps>(
     },
     ref
   ) => {
-    const closeModal = () => {
+    const prevHide = useMemo(() => state.hide, [state])
+
+    state.hide = useCallback(() => {
```
mleralec, line 71: maybe it's better to put that directly in the `useModalState` ?

Isaacmaamouche: The custom onClose [is passed as a prop](https://github.com/WTTJ/welcome-ui/blob/fc1db748468c974130d7f7224e2f37f4b246b695/packages/Modal/src/index.tsx#:~:text=export%20interface%20ModalOptions,%2C%20ModalOptions%3E)
Should we change it to a hook option and remove it form the modalProps ?

Calvein: I tend to agree with Mick, monkey patching the state & mutating the state isn't great imo

packages/Modal/src/index.tsx
```diff
@@ -66,25 +66,26 @@ const ModalComponent = forwardRef<'div', ModalProps>(
...
-            hide={closeModal}
             hideOnClickOutside={hideOnClickOutside}
             preventBodyScroll={false}
             ref={ref}
             size={size}
             {...state}
             {...rest}
           >
-            <CloseElement onClick={closeModal} />
+            <CloseElement onClick={() => state.hide()} />
```
anucreative, line 88:
```suggestion
            <CloseElement onClick={state.hide} />
```

anucreative (commented):
Très petit commentaire :)

packages/Modal/src/index.tsx
Isaacmaamouche: It triggers an eslint error "Handler function for onClick prop key must be a camelCase name beginning with 'handle' only"

packages/Modal/src/index.tsx
```diff
@@ -43,12 +42,23 @@ export type ModalInitialState = DialogInitialState & {
...
   const dialogState = useDialogState({ animated: true, visible: visible || open, ...restOptions })
 
+  const prevHide = useMemo(() => dialogState.hide, [dialogState])
+
+  dialogState.hide = useCallback(() => {
+    onClose?.()
+    prevHide()
+  }, [prevHide, onClose])
+
   return { open: dialogState.visible, ...dialogState }
```
mleralec, line 62: un truc comme ça devrait faire le taff ?:
```tsx
  return {
    ...dialogState,
    open: dialogState.visible,
    hide: () => {
      dialogState.hide()
      onClose?.()
    }
  }
```

ça évite de modifier `dialogState` et on modifie juste notre "`modalState`"

mleralec (approved):

PR #2068, opened 2023-05-17
Fix: rounding of swiper's nextPage calculation

Isaacmaamouche:
The recently applied fix: https://github.com/WTTJ/welcome-ui/pull/2052/files passed manual testing on simple configured swipers but it introduced a bug on swipers with multiple slides per view and at least 3 pages. 

Fix: properly round the whole result of nextPage's calculation. 

https://github.com/WTTJ/welcome-ui/assets/85453899/26fdec4c-d117-4569-8db5-d915ef7dbc03


theo-mesnil:
Hello, possible de commencer à mettre un test sur le composant pour ce soucis ?

PR #2083, opened 2023-06-12
fix: uniformize how drawer trigger is exported and avoid TS error

P1X3L:
Drawer now exposes it's Trigger component the same way a Modal does.

It avoids TS errors with the `as` prop ie :
image


theo-mesnil:
Replaced by #2087

PR #2087, opened 2023-06-14
fix: drawer trigger "as component" props

mleralec:
related to: https://github.com/WTTJ/welcome-ui/pull/2083

The `Drawer.Trigger` throw errors when you add props of the "as" component:
```tsx
<Drawer.Trigger as={Button} variant="primary" state={state}>...</Drawer.Trigger>
```
Here, the prop "variant" will throw an error, if you remove it, it works.

This is because the `Drawer.Trigger` component doesn't use our custom `forwardRef` type (this type inject props of the as component in the current component, _here add the props of the `Button` in the `Drawer.Trigger`_)

### DIFF
--- a/packages/Drawer/src/index.tsx
+++ b/packages/Drawer/src/index.tsx
@@ -3,7 +3,6 @@ import {
   Dialog,
   DialogBackdropProps,
   DialogDisclosure,
-  DialogDisclosureProps,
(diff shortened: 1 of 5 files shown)


theo-mesnil (approved):
👏🏻

PR #2205, opened 2023-08-21
Use theme spaces for DropdownMenu gutter

Calvein:
Rather than forcing the user to set a number for the number, I've added the spaces coming from the theme in addition.

image

@Jademhm & @aurelienalletru we have a spacing (the gutter) of 10 between the trigger & the dropdown by default, would you rather have 8 (sm) or 12 (md)?


theo-mesnil (commented):
I think we don't want to add another value from theme, 10px is the spacing from the specs by default and we need all to use it to be consistent

Calvein:
> @Calvein but i think we need to get default space from theme and not with a number

That was a suggestion but I didn't look at the specs about the 10px, I thought it was an old value that we kept 👀

I've added the gutter to the theme and I added it as a default in https://github.com/WTTJ/welcome-ui/pull/2205/commits/fd3e3825aa9ad281bf92319a46e2719a70aebb10

theo-mesnil:
Capture d’écran 2023-08-22 à 14 30 07

@Calvein i check the figma and it's now `xs`

packages/DropdownMenu/src/index.tsx
```diff
@@ -1,31 +1,58 @@
...
+    | 'sm'
+    | 'md'
+    | 'lg'
+    | 'xl'
+    | 'xxl'
+    | '3xl'
+    | '4xl'
+    | '5xl'
+    | '6xl'
+    | '7xl'
```
theo-mesnil, line 26: Get the space theme types instead or rewrite it

Calvein: I tried to get the theme with `ThemeSpace` from `@xstyled/styled-components` but its type is `string | number | true | symbol | {} ` rather than this list, so I had to add `@welcome-ui/core` to get it.

packages/DropdownMenu/src/index.tsx
```diff
@@ -1,31 +1,46 @@
...
 import { Arrow } from './Arrow'
 import { Item } from './Item'
 import { Separator } from './Separator'
 import * as S from './styles'
 
-export interface DropdownMenuOptions extends Ariakit.MenuProps {
+export interface DropdownMenuOptions extends Omit<Ariakit.MenuProps, 'gutter'> {
   /** add custom props from styled system on DropdownMenu inner */
   innerProps?: WuiProps
+  gutter?: keyof WuiTheme['space'] | number
```
theo-mesnil, line 15:
```suggestion
/** default is 4px (xs) */
  gutter?: keyof WuiTheme['space'] | number
```

packages/DropdownMenu/src/index.tsx
```diff
@@ -1,31 +1,46 @@
...
 
 const DropdownMenuComponent = forwardRef<'div', DropdownMenuProps>(
-  ({ children, dataTestId, innerProps = {}, store, gutter = 10, ...rest }, ref) => {
+  ({ children, dataTestId, innerProps = {}, store, gutter = 'xs', ...rest }, ref) => {
+    const theme = useTheme()
     const arrowElement = store.useState('arrowElement')
     const isOpen = store.useState('open')
 
+    let parsedGutter = gutter
+    if (typeof parsedGutter === 'string') {
```
theo-mesnil, line 27: Add test for this cases please

Calvein: Done and keen to get your feedback on [the commit](https://github.com/WTTJ/welcome-ui/pull/2205/commits/4b06a1c0c9abe81f8b8d0e78ab33988cf9c80c3f#diff-355b56cfd7481bba9c9332e59b81e99caa4794311322945ac7cd1e8724820b36R27)

theo-mesnil:
@Calvein need rebase :)

Calvein:
All good 🚀

theo-mesnil (approved):

PR #2292, opened 2023-09-21
fix: swiper custom button background

theo-mesnil:
We need to override the ghost button with important because on our case the background is white, and styled component doing wrong override here so we need to add important to be sure is overrided.


anucreative (approved):
🤮 😂

PR #2338, opened 2023-10-10
fix(tooltip): always visible after hover

mleralec:
Workaround for this issue: https://github.com/ariakit/ariakit/issues/2929

### DIFF
--- a/packages/Tooltip/src/styles.ts
+++ b/packages/Tooltip/src/styles.ts
@@ -35,11 +35,10 @@ const fadeInStyle = css`
 type FadeIn = {
   placement?: TooltipProps['placement']
   fixed?: boolean
-  isOpen?: boolean
 }
 
 export const FadeIn = styled.div<FadeIn>(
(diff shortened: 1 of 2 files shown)


theo-mesnil (approved):

packages/Tooltip/src/styles.ts
```diff
@@ -55,11 +54,6 @@ export const FadeIn = styled.div<FadeIn>(
     [data-enter] & {
       ${fadeInStyle}
     }
```
diegohaz, line 56: I think the bug here is that the CSS transition is not being applied to the tooltip element. The tooltip doesn't hide because it's still waiting for a CSS transition/animation to complete.

You can apply the transition directly to the tooltip element or manually call `tooltip.stopAnimation()`:

```jsx
<S.FadeIn onTransitionEnd={tooltip.stopAnimation}>
```

mleralec: @diegohaz Thanks for your answer, I'll check that 🙏🏻

Issue #2341, opened 2023-10-16
[Modal] isn't visible when we update the state during the opening


mleralec:
### [Design] For which Figma library?

None

### For which component/foundations?

Modal

### Prioritization

High 🌋

### Current Behavior

When we open the modal and updating a state at the same time, the modal doesn't open (but the backdrop yes).

The `data-enter` attribute is forwarded to the **Backdrop** component but not to the **Modal** and the opacity remains at 0.

Reproduce:
1. open https://codesandbox.io/s/bug-modal-while-updating-state-y544rd?file=/src/App.tsx
2. click on the button to open the modal
3. the backdrop is visible but not the modal

### Expected Behavior

The `data-enter` attribute is forwarded to the Modal component too.

### Screenshots / Videos

https://codesandbox.io/s/bug-modal-while-updating-state-y544rd?file=/src/App.tsx

### [Tech] Welcome UI Version

5.0.5

### [Tech] Browser

Google Chrome 118.0

### [Tech] Operating System

MacOS

PR #2342, opened 2023-10-16
fix(modal): not visible while updating state on open

mleralec:
Fix https://github.com/WTTJ/welcome-ui/issues/2341

The first option is to remove the `onClose` prop from the `useModal` and render the same hook as the drawer:
```tsx
export function useModal(options?: UseModalProps): UseModal {
  return Ariakit.useDialogStore({ animated: true, ..options })
}
```

The second option is to wrap the render of the hook with a `useMemo`, I decided to choose this one for prevent breaking changes, but I'm open to discussing it.

(I prefer the first option, but I don't know if the `onClose` prop is used by someone?)

### DIFF
--- a/packages/Modal/src/index.tsx
+++ b/packages/Modal/src/index.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react'
+import React from 'react'
 import * as Ariakit from '@ariakit/react'
 import { BoxProps } from '@welcome-ui/box'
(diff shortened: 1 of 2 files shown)


diegohaz:
You can also use the [`setOpen`](https://ariakit.org/reference/use-dialog-store#setopen) callback on [`useDialogStore`](https://ariakit.org/reference/use-dialog-store):

```js
useDialogStore({
  setOpen(open) {
    if (!open) {
      onClose?.();
    }
  }
})
```

PR #2358, opened 2023-11-07
fix: hide unwanted aria messages for date/time pickers

### LINKED ISSUE #2305: [TimePicker] the aria live message is always shown
### For which component?

TimePicker

### Description

A message containing the selected date is always shown above the datepicker when we select a date:

https://github.com/WTTJ/welcome-ui/assets/36373219/f2f1f195-6945-43c0-a8eb-80f9c7939655

maybe related to https://github.com/WTTJ/welcome-ui/issues/2171

### Welcome UI Version
5.0.0

simon-hv:
Fixes #2305 

Hide unwanted aria message for DatePicker and TimePicker

screenshot_2023-11-06_at_10 47 01

Screenshot 2023-11-07 at 14 21 13

### DIFF
--- a/packages/DateTimePickerCommon/src/CustomPopper.tsx
+++ b/packages/DateTimePickerCommon/src/CustomPopper.tsx
@@ -3,6 +3,7 @@ import styled, { css, CSSObject, th } from '@xstyled/styled-components'
 import { cardStyles } from '@welcome-ui/utils'
 
 import { datePickerStyles } from './datePickerStyles'
+import { fixAriaMessageStyle } from './styles'
 
 export const CustomPopper = ({
   children,
@@ -21,6 +22,7 @@ export const CustomPopper = ({
 const StyledCustomPopper = styled.div(
   ({ popperStyles }: { popperStyles: CSSObject }) => css`
     ${datePickerStyles};
+    ${fixAriaMessageStyle};
     .react-datepicker-popper {
       ${popperStyles};
     }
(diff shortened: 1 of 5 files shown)


packages/DateTimePickerCommon/src/CustomPopper.tsx
```diff
@@ -21,6 +22,7 @@ export const CustomPopper = ({
 const StyledCustomPopper = styled.div(
   ({ popperStyles }: { popperStyles: CSSObject }) => css`
     ${datePickerStyles};
+    ${FixAriaMessageStyle};
```
mleralec, line 25: doesn't work if you apply the style just right here?
I think we should not use a global style in this case because we can style the `Popper` directly

simon-hv: It works only for the `DatePicker` because the aria message is displayed inside the `Popper`. 
But for the`TimePicker`, it doesn't work because the message is displayed above the input:
Screenshot 2023-11-07 at 14 46 19

mleralec: maybe we can use this solution for the `DatePicker` and for the `TimePicker`, use a wrapper that includes this style?

simon-hv: ✅

packages/DateTimePickerCommon/src/styles.ts
```diff
@@ -74,3 +74,17 @@ export const Selects = styled.div`
     margin-right: sm;
   }
 `
+
+export const FixAriaMessageStyle = css`
```
mleralec, line 78: maybe `fixAriaMessageStyle` is better because it's not a component?

simon-hv: you're right 👍

mleralec (approved):

PR #2398, opened 2024-02-05
Fix <Modal.Footer> type

Calvein:
There is no reason for them to be string, and we already pass components in some internal projects, so `ReactNode` should be better.

### DIFF
--- a/packages/Modal/src/Footer.tsx
+++ b/packages/Modal/src/Footer.tsx
@@ -7,8 +7,8 @@ import * as S from './styles'
 
 export interface FooterOptions {
   information?: {
-    title: string
-    subtitle: string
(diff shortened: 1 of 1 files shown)


theo-mesnil:
@Calvein it's for translation component ?

Calvein:
Not in the cases I just fixed, it was something like:
```tsx
<Box>
  <FormattedMessage />
  <Link />
</Box>
```

theo-mesnil:
@Calvein it's a custom footer, i think you need a custom footer too ? We have done this to avoid custom style on footer

Calvein:
I think you're right, we have a weird mix which is, I'm guessing, due to the migration to the v5, I'll fix it, thanks!

PR #2422, opened 2024-02-29
fix: a11y on modal header

Leiksa:


docs/pages/components/modal.mdx
```diff
@@ -123,7 +123,7 @@ function() {
       </Modal.Trigger>
       <Modal store={modal} ariaLabel="example">
         <Modal.Content store={modal}>
-          <Modal.Header title={title} subtitle={subtitle} icon={icon}/>
+          <Modal.Header title={<div>hello</div>} subtitle={subtitle} icon={icon}/>
```
theo-mesnil, line 126: Remove this example :) we need to avoid maximum the use of other than string ^^

Leiksa: oops, just forgot to remove when i test it 🙈

theo-mesnil (approved):

PR #2423, opened 2024-03-07
chore: upgrade ariakit to v0.4.3 and upgrade test with userEvent

theo-mesnil:


packages/Modal/src/index.tsx
```diff
@@ -112,8 +112,15 @@ const Cover: React.FC<ShapeProps> = props => {
 
 type TriggerProps = { store: Ariakit.DialogStore; children: React.ReactNode; as?: As }
 
-export const Trigger = forwardRef<'button', TriggerProps>(({ as, store, ...rest }, ref) => {
-  return <Ariakit.DialogDisclosure as={as} ref={ref} store={store} {...rest} />
+export const Trigger = forwardRef<'button', TriggerProps>(({ as: As, store, ...rest }, ref) => {
+  return (
+    <Ariakit.DialogDisclosure
+      ref={ref}
+      render={As ? props => <As {...props} ref={ref} /> : undefined}
```
mleralec, line 119: `render={as}` isn't enough? 🤔

mleralec: I think props/ref are forwarded to the `as` component but maybe not 🤷🏻‍♂️

mleralec: or not :grin: https://ariakit.org/guide/composition

mleralec (approved):

PR #2436, opened 2024-03-27
Core upgrade xstyled styled components v6

P1X3L:
* use xstyled fork with styled-components v6.1.8
* upgrade styled-components peer dep to v6.1.8


README.md
```diff
@@ -19,7 +19,7 @@ Here you'll find all the core components you need to create a delightful webapp.
 1 - Install the **peer dependencies** listed below:
 
 ```bash
-yarn add @xstyled/styled-components react styled-components
+yarn add @wttj/xstyled-styled-components react styled-components
```
guillaumewttj, line 22: is it public ?

packages/System/src/index.tsx
```diff
@@ -57,11 +56,11 @@ const WRAPPER_PROPS = Object.freeze([
 ])
 
 /**
- * @deprecated use system from @xstyled/syled-components instead
+ * @deprecated use system from @wttj/xstyled-syled-components instead
  */
 export const system = compose<WuiSystemProps>(...SYSTEM_PROPS)
 /**
```
guillaumewttj, line 62: since we've stopped using it in favor of *Box components; should we remove it from the exports ?

packages/Text/src/styles.ts
```diff
@@ -39,6 +60,9 @@ export const Text = styled.p<TextOptions>(({ lines, variant }) => {
       ${system};
     }
 
-    ${system};
+    ${lines &&
+    lines !== Infinity &&
+    Object.keys(lineHeightHeadingsFixer).includes(variant) &&
+    fixLineHeightStyles(variant)};
```
guillaumewttj, line 66: i dont understand how this is related to the upgrade ?

packages/Utils/src/get-max-width-height.ts
```diff
@@ -1,6 +1,20 @@
...
+    (value: SystemProp<string | number, Theme>) => number
+  > = {
+    string: (value: string) => parseInt(value, 10),
+    number: (value: number): number => value,
+  }
+  const widthNumber =
+    typeToValueExtractor[typeof width as keyof typeof typeToValueExtractor]?.(width) || 0
+  const heightNumber =
+    typeToValueExtractor[typeof width as keyof typeof typeToValueExtractor]?.(height) || 0
+  const diff = widthNumber - heightNumber
```
guillaumewttj, line 18: i dont see the problem that has been solved here, could you clarify it ?

README.md
theo-mesnil: Not yet but it's will be discuss tomorrow :)

packages/Text/src/styles.ts
P1X3L: It has nothing to do with the upgrade indeed 🙈
Just couldn't keep up with those g, y, q, j and p being cut

packages/System/src/index.tsx
P1X3L: We still need it to be able to override style props in particular use cases

packages/Utils/src/get-max-width-height.ts
P1X3L: the `width` and`height` arguments were typed as `string` but have been enriched after the upgrade of xstyled & styled-components resulting in type errors

theo-mesnil:
@P1X3L on peut merger avec la PR #2466 ?

theo-mesnil:
Replaced by #2466

PR #2455, opened 2024-06-07
chore: allow shape prop to use object notation with breakpoints as keys

P1X3L:
To avoid Cumulative Layout Shift, we have spotted that we could leverage the usage of the object notation on Button's `shape` prop (like we do on style props)
(more info here 👉 https://www.debugbear.com/project/865/pageLoad/6293/overview?metric=cumulativeLayoutShift focus on the share button)

Until now, button's `shape` attribute had the type `'circle' | square`

It can now be `circle | square | default | Record<BREAKPOINT KEYS HERE, ('circle' | 'square' | 'default')>`

We have an app using `Button` with the `shape` prop being dynamically set with JS to `circle` depending on breakpoints.
Unfortunately, these dynamic styles are not working on SSR so it is better to use css to do so.

### DIFF
--- a/docs/pages/components/button.mdx
+++ b/docs/pages/components/button.mdx
@@ -28,6 +28,7 @@ Built with [Ariakit](https://ariakit.org/components/button) for a better accessi
 <Button variant="secondary">Secondary</Button>
 <Button variant="tertiary">Tertiary</Button>
 <Button variant="ghost">Ghost</Button>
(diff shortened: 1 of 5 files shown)


mleralec:
From my point of view, this code adds a lot of complexity for a very small use case (especially since we maybe want to remove the "shape" prop in the future?) I like to keep WUI as simple as possible 🤷🏻‍♂️

P1X3L:
Ok, I see your point, I'll make the changes where I found them.
I'm closing the issue then.

PR #2466, opened 2024-07-02
chore: upgrade styled-components/xstyled

mleralec:


P1X3L (approved):
Nice work @mleralec thank you!

Is there a reason why you keep using system / shouldForwardProp where we could use `xBox` all over?

mleralec:
hard rebase, new PR https://github.com/WTTJ/welcome-ui/pull/2524

PR #2256, opened 2023-09-13
[S][SRE-458] remove netlify and push on own stack the website

ClemWttJ:
**Link to issue  #2255** 

▶️ **Waiting after release for complete migration**

Issue #2255, opened 2023-09-13
[Infra] Remove Netlify deployment


ClemWttJ:
### Preprod
⚪   Url new format `custom_branch.preprod.welcome-ui.com`

- Update in pipelines to remove netlify commands 
- Extract branch and format it for circleci and urls standards

### Production
⚪  Same url `www.welcome-ui.com`

- Update in pipelines to remove netlify commands 

---
### Bug/Work in progress 🐛 

1.  Site not completely static, can't access specific component directly with a link [e.g](https://sre-458-migration-welcome-ui-sur-k-8-s.preprod.welcome-ui.com/components/swiper)

### Solution

1. ▶️  Rework documentation as full static
    ▶️  Update lambda to serve multiples path (not sure if possible)

ClemWttJ:
**Issue stalled, waiting to complete solution** 
### Solution
1. Rework documentation as full static ▶️  link to #2164

PR #2483, opened 2024-07-17
[SRE-78] migration WUI

ClemWttJ:
_Old [PR#2256](https://github.com/WTTJ/welcome-ui/pull/2256)_ 

Link to issue https://github.com/WTTJ/welcome-ui/issues/2255
Rebase was too messy so it was best to start over and correctly update the pipeline !


.circleci/config.yml
```diff
@@ -129,79 +149,94 @@ jobs:
...
+          aws_access_key_id: AWS_ACCESS_KEY_ID
+          aws_secret_access_key: AWS_SECRET_ACCESS_KEY
+          region: ${AWS_REGION}
+      - run:
+          name: Process branch name
+          command: |
+            echo "export PROCESSED_BRANCH=$(echo "$CIRCLE_BRANCH" | sed "s/[^[:alnum:]-]/-/g" | tr '[:upper:]' '[:lower:]' | cut -c -63)" >> $BASH_ENV
+      - aws-s3/sync:
+          from: website/out
+          to: s3://welcome-ui/preprod/${PROCESSED_BRANCH}
```
mleralec, line 169: Is it possible to replace `preprod` by `preview` on the URL of the website? 👀

ClemWttJ: Hello, yes it's possible but:
- we need to change the record (host)
- since we change the host, we need to change the lambda behind the dynamic websites
- we also need to change the path of the bucket

What's bother me a little is that for all preprod stage we are using `preprod` in our records, and it means changing the logic only for welcome-ui 😄
But I don't have a strong opinion, it's not critical ! What do you think @pauldn-wttj ?

pauldn-wttj: That's up to you @mleralec.

`welcome-ui` is not exactly following the same lifecycle as our other applications since it's an open source design system so I understand we don't have to have the rules. If you prefer to have `preview` instead of `preprod`, I'd say now is the time to make the switch. :slightly_smiling_face:

mleralec: Thanks for your answers @ClemWttJ @pauldn-wttj !
cc @theo-mesnil what do you think about that?

theo-mesnil: I prefer preview because is not really a preprod :) if it's possible ❤️

ClemWttJ: Ok on it then 😄

ClemWttJ: Good for us :+1:

pauldn-wttj (approved):

mleralec (approved):
🔥🔥🔥

theo-mesnil (approved):

PR #2509, opened 2024-08-06
[SUPPORT] correct pipeline font

ClemWttJ:
The update of the orb `aws-s3` caused a wrong command interpretation ▶️ metadatas were added to folder name instead of object properties


theo-mesnil (approved):
🚀

PR #2513, opened 2024-08-16
feat: allow to display custom file name in the preview of a FileDrop

simon-hv:
Add a new `fileName` prop to the `FileDrop` component, allowing users to display a custom file name in the preview instead of the default name extracted from the URL.  
This is useful when the actual file name is too complex or not representative of the file's content.


theo-mesnil:
I think just fileName can be enough for the name of property

simon-hv:
> I think just fileName can be enough for the name of property

You're right, I just made the change @theo-mesnil

theo-mesnil (approved):

PR #2520, opened 2024-08-23
chore: upgrade dependencies for minors and fixes

theo-mesnil:


mleralec (approved):

.circleci/config.yml
```diff
@@ -7,17 +7,17 @@ orbs:
...
-    default: 'cimg/node:18.17.1'
+    default: 'cimg/node:20.17.0'
 
 executors:
   nodejs:
     docker:
       - image: << pipeline.parameters.nodejs_base_image >>
   docker_builder:
-    resource_class: medium
+    resource_class: large
```
guillaumewttj, line 20: We should not do that for every container in the workflow. Most jobs should run on small ressource_class, in some rare cases, you need a larger machine ! (so if you need more resource for a job, apply a custom resource class to it)

theo-mesnil:
Will be split in multiples PR

PR #2561, opened 2024-10-17
[WUI-45] fix: avoid welcome-font cut letters

P1X3L:


packages/Text/src/styles.ts
```diff
@@ -11,6 +11,15 @@ const MOBILE_VARIANTS = {
...
 }
+const headingsLineHeightFixer = {
+  h0: 0.825,
+  h1: 0.75,
+  h2: 0.825,
+  h3: 1.075,
+  h4: 1,
+  h5: 0.9,
+  h6: 0.9,
+}
```
theo-mesnil, line 22: Maybe we can add all of this on typography core theme ?

P1X3L: I just wanted to keep all the ugly fixes in the same place until we have a correct font 🫠

PR #2562, opened 2024-10-21
feat: rebranding with new colors and logos [Version 6]

theo-mesnil:
Frame 4


packages/Avatar/src/utils.ts
```diff
@@ -1,9 +1,21 @@
...
+  3: 'orange',
+  4: 'green',
+  5: 'pink',
+  6: 'violet',
+}
 
 export function getSeededColor(colors: Theme['colors'], seed = ''): string {
-  const colorKeys = Object.keys(colors).filter(color => color.startsWith('sub-'))
+  const colorKeys = Object.keys(colors).filter(color => color.startsWith('secondary-'))
   const subColorNumber = (seed.length % colorKeys.length) + 1
```
P1X3L, line 17:
```suggestion
  const subColorIndex = seed.length % colorKeys.length
```

packages/Avatar/src/utils.ts
```diff
@@ -1,9 +1,21 @@
...
-  const colorsIndex = `sub-${subColorNumber}` as keyof Theme['colors']
+  const colorsIndex = `secondary-${subColorByNumber[subColorNumber]}` as keyof Theme['colors']
```
P1X3L, line 18:
```suggestion
  const secondaryColor = colorKeys[colorIndex] as keyof Theme['colors']
```

packages/Avatar/src/utils.ts
```diff
@@ -1,9 +1,21 @@
...
 
   return colors[colorsIndex]
```
P1X3L, line 20:
```suggestion
  return secondaryColor
```

packages/Avatar/src/utils.ts
```diff
@@ -1,9 +1,21 @@
 import { Theme } from '@xstyled/styled-components'
+import { ThemeSecondaryColors } from '@welcome-ui/core'
+
+const subColorByNumber: {
```
P1X3L, line 4: You can manage to avoid naming colors here for `getSeededColor` can work without it I think.

packages/Alert/src/index.tsx
```diff
@@ -1,15 +1,15 @@
...
 import { Button, ButtonProps } from '@welcome-ui/button'
 import { CreateWuiProps, forwardRef } from '@welcome-ui/system'
-import { Variant, VariantIcon } from '@welcome-ui/variant-icon'
+import { VariantIcon } from '@welcome-ui/variant-icon'
 
 import * as S from './styles'
 import { Title } from './Title'
 
 export type Size = 'sm' | 'md'
+export type Variant = 'danger' | 'success' | 'warning' | 'info' | 'default' | 'beige'
```
anucreative, line 12: Do we still use `'info'` ?

theo-mesnil: Only on Alert / Growl / Snarkbar / Tag :)

PR #2572, opened 2024-10-28
Chore(Select): use pointer cursor when hovering select options [WUI-50]

Isaacmaamouche:
[WUI-50: Add a pointer cursor to Select dropdown](https://linear.app/wttj/issue/WUI-50/add-a-pointer-cursor-to-select-dropdown)

select pointer cursor after


theo-mesnil (approved):

theo-mesnil (changes requested):
We need to have also on parent input the cursoir pointer

theo-mesnil (approved):

PR #2591, opened 2024-11-05
feat: update alert component

stevenpersia:
https://github.com/user-attachments/assets/93f5578e-4a82-4a52-b516-197c3cb26617

### DIFF
--- a/packages/Alert/src/index.tsx
+++ b/packages/Alert/src/index.tsx
@@ -31,6 +31,17 @@ type CloneActionsReturns = React.ReactElement<
   string | React.JSXElementConstructor<AlertProps>
 >
 
+const LAYOUT: { bottom: BoxProps; right: BoxProps } = {
+  bottom: {
+    alignItems: 'flex-start',
+    flexDirection: 'column',
+  },
+  right: {
+    alignItems: { _: 'flex-start', md: 'center' },
+    flexDirection: { _: 'column', md: 'row' },
+  },
+}
+
 const AlertComponent = forwardRef<'div', AlertProps>(
   (
     {
@@ -58,8 +69,8 @@ const AlertComponent = forwardRef<'div', AlertProps>(
 
     // Handle clone actions recursively in case of multiple buttons
     const cloneActions = (child: React.ReactElement<AlertProps>): CloneActionsReturns => {
-      if (child.type === AlertPrimaryAction) return cloneElement(child, { size })
-      if (child.type === AlertSecondaryAction) return cloneElement(child, { size })
+      if (child.type === AlertButton) return cloneElement(child, { size })
+      if (child.type === AlertSecondaryButton) return cloneElement(child, { size })
 
       if (child.props?.children) {
         return cloneElement(child, {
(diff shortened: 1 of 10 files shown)


packages/Alert/src/index.tsx
```diff
@@ -44,13 +51,41 @@ const AlertComponent = forwardRef<'div', AlertProps>(
...
+    const layout: { bottom: BoxProps; right: BoxProps } = {
+      bottom: {
+        alignItems: 'flex-start',
+        flexDirection: 'column',
+      },
+      right: {
+        alignItems: { _: 'flex-start', md: 'center' },
+        flexDirection: { _: 'column', md: 'row' },
+      },
+    }
```
theo-mesnil, line 87: Can we move to top before the component ?

packages/Alert/src/index.tsx
```diff
@@ -74,34 +109,36 @@ const AlertComponent = forwardRef<'div', AlertProps>(
...
-              <div>{content}</div>
-              {cta}
-            </Box>
-          ) : (
-            content
-          )}
+        <Box flex={1}>
+          <Box display="flex" gap="md" justifyContent="space-between" {...layout[ctaPosition]}>
+            <Box flex={1}>{content}</Box>
+
```
theo-mesnil, line 118: remove empty line

packages/Alert/src/index.tsx
```diff
@@ -74,34 +109,36 @@ const AlertComponent = forwardRef<'div', AlertProps>(
...
+const AlertSecondaryAction = forwardRef<'button', Omit<ButtonProps, 'size' | 'variant'>>(
+  (props, ref) => <Button flexShrink={0} ref={ref} w="fit-content" {...props} variant="tertiary" />
+)
 
-export const Alert = Object.assign(AlertComponent, { Title, Button: AlertButton })
+export const Alert = Object.assign(AlertComponent, {
+  Title,
+  PrimaryAction: AlertPrimaryAction,
+  SecondaryAction: AlertSecondaryAction,
```
theo-mesnil, line 143: I think this is a breaking change, maybe we can for now export `Button: AlertPrimaryAction` ? With a deprecated message ?

theo-mesnil: Or rename` Button / SecondaryButton` ?

packages/Alert/src/index.tsx
stevenpersia: You're right 👍 Fixed here : b3e2450.

packages/Alert/src/index.tsx
stevenpersia: Done here b3e2450.

packages/Alert/src/index.tsx
stevenpersia: Nice catch! Yes, I tried to reduce the breaking changes on this component, but I forgot about that. You are absolutely right. I kept the current button and put the second one (see b3e2450). So, maybe we don't need deprecated anymore, it's just a styling update?

theo-mesnil (approved):

theo-mesnil (approved):

PR #2607, opened 2024-11-12
docs: change migration after return of primary variant on Tag

theo-mesnil:


website/build-app/pages/foundations/migration.mdx
```diff
@@ -266,12 +266,7 @@ We removed `shape` property on Badge component, now you will only have rounded b
 
 ### Tag
 
-We removed variant `primary` and `secondary` and replace the `sub-*` variants. To have the primary color please use Badge component instead.
-
-```diff
-- <Tag variant="primary">Name</Tag>
-+ <Badge>Name</Badge>
-```
+We removed variant `secondary` and replace the `sub-*` variants. To have the primary color please use Badge component instead.
```
anucreative, line 269:
```suggestion
We removed the `secondary` variant and replaced the `sub-*` variants. To use the `secondary` color please use the `Badge` component instead.
````

anucreative (approved):

PR #2611, opened 2024-11-15
fix(alert): align title when no description

RobelTekle:
Fix elements alignement when there is not Title
**Before**
Capture d’écran 2024-11-15 à 10 53 00


cnairi (approved):

theo-mesnil:
We don't want to have an alert with only a title, we always have a description

RobelTekle:
> We don't want to have an alert with only a title, we always have a description

👋 Hello @theo-mesnil, I agree with you but the problem is that we already have this cases in our projets. I suggest that we handle this in any case (no conflict with the current design) because the process of creating new wordings for each of theses cases could really slow down the migration.

theo-mesnil:
@RobelTekle i'll come back to you i'll check with Jade but i think you juste need to remove <Alert.Title> and only put your wording in Alert

PR #2642, opened 2025-01-07
feat(Text): add wordBreak property

cnairi:
# DESCRIPTION

We want to implement the wordBreak property on the Text component to avoid breaking the display of this component in our apps.

## LINEAR TICKET

https://linear.app/wttj/issue/WUI-70/components-do-not-support-word-break-or-word-wrap-properties

# HOW TO TEST

- You need to run `yarn start` on your project locally
- Then go to `http://localhost:3020/components/text`
- See you now have a `wordBreak` example in the doc, with long words that need to be broken

# SCREENSHOTS / SCREEN RECORDINGS

Capture d’écran 2025-01-07 à 12 05 57

### DIFF
--- a/packages/Text/docs/examples/wordBreak.tsx
+++ b/packages/Text/docs/examples/wordBreak.tsx
@@ -1,0 +1,22 @@
+import * as React from 'react'
+import { Text } from '@welcome-ui/text'
+
+const Example = () => {
+  return (
+    <>
+      <Text wordBreak="break-all">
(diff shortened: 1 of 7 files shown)


Isaacmaamouche (approved):

theo-mesnil:
@cnairi i think we need to break work everytime and let the possibility to change it ? And you need to rebase sorrry :)

PR #2684, opened 2025-02-14
Chore(Types): move duplicate types to theme [WUI-110]

Isaacmaamouche:
### DESCRIPTION
- Removes duplicate type and import them from theme when they're used in multiple places
- Exports a new type `ThemeColorTokens` for all our colors name
- Introduces a new type util **LiteralUnion<T>**
`LiteralUnion<T, U = string> = T | (U & Record<never, never>)`

Preserve type intellisence and autocomplete on union with string
```typescript
// The following type:
type color = 'red' | 'blue' | 'green' | string
// would be interpreted as 
type color = string
// and provide no autocompletion
``` 

With the new util `LiteralUnion<T, U>` you can pass your type as T and you union target as U (default to string)
```typescript
// The following type:
type color = LiteralUnion<'red' | 'blue' | 'green'>
// would be interpreted as 
type color = 'red' | 'blue' | 'green'
// While still allowing any literal string without error
 * const bg: color = 'yellow'
 * const bg: color = '#FFF'
``` 
Works with number as well
```typescript
type LoarderSize = 'xs' | 'sm' | 'md' | 'lg'
export type Size = LiteralUnion<LoarderSize, string | number>
```


lib/src/components/Loader/theme.ts
```diff
@@ -1,11 +1,10 @@
...
-export type ThemeLoaders = {
-  lg: string
-  md: string
-  sm: string
-  xs: string
-}
+type LoarderSize = 'xs' | 'sm' | 'md' | 'lg'
+export type Size = LiteralUnion<LoarderSize, string | number>
+
+export type ThemeLoaders = Record<LoarderSize, string>
```
anucreative, line 7:
```suggestion
type LoaderSize = 'xs' | 'sm' | 'md' | 'lg'
export type Size = LiteralUnion<LoaderSize, string | number>

export type ThemeLoaders = Record<LoaderSize, string>
```

Sorry to be so pedantic 🤓

lib/src/components/Loader/theme.ts
```diff
@@ -1,11 +1,10 @@
...
+import { LiteralUnion } from '@/utils'
 
-export type ThemeLoaders = {
-  lg: string
-  md: string
-  sm: string
-  xs: string
-}
+type LoarderSize = 'xs' | 'sm' | 'md' | 'lg'
+export type Size = LiteralUnion<LoarderSize, string | number>
```
cnairi, line 5: It's nothing but is it Loarder or Loader here in `LoarderSize` ?

anucreative (approved):
Très propre ! 👏🏼

lib/src/components/Toggle/theme.ts
```diff
@@ -6,6 +6,7 @@ import { ThemeValues } from '@/theme'
 const XS_LINE_HEIGHT_ADJUSTMENTS = '0.32rem !important'
 const SM_LINE_HEIGHT_ADJUSTMENTS = '0.12rem ! important'
 
+export type Size = 'xs' | 'sm' | 'md'
 type State = 'default' | 'checked' | 'disabled' | 'sizes'
```
anucreative, line 10: Nothing to do with your PR but what is `sizes`  as a State. Sounds very strange and doesn't seem to fit with `'default' | 'checked' | 'disabled'` 🤔

lib/src/components/Modal/theme.ts
```diff
@@ -2,7 +2,7 @@ import { CSSObject } from '@xstyled/styled-components'
 
 import { ThemeValues } from '@/theme'
 
-type Sizes = 'xs' | 'sm' | 'md' | 'lg' | 'auto'
+export type Size = 'xs' | 'sm' | 'md' | 'lg' | 'auto'
```
anucreative, line 5: 👍🏼

lib/src/components/Loader/theme.ts
Isaacmaamouche: Supposed to be Loader, i can't write 😭

lib/src/components/Loader/theme.ts
Isaacmaamouche: Good catch! I should go back to school 😭

lib/src/components/Toggle/theme.ts
Isaacmaamouche: Agreed
@theo-mesnil is it okay if I break it down like this: 

```typescript
export type Size = 'xs' | 'sm' | 'md'
type State = 'default' | 'checked' | 'disabled'

export type ThemeToggles = {
  after: Record<State | 'sizes', CSSObject>
  icon: Record<'sizes' | 'position', CSSObject>
  item: Record<State | 'sizes', CSSObject>
}
```

theo-mesnil: Yes let's go :)

theo-mesnil (approved):
👏🏻

PR #2706, opened 2025-03-13
chore: rework linters with new configuration

theo-mesnil:


lib/src/components/ButtonGroup/index.tsx
```diff
@@ -1,12 +1,10 @@
 import React, { Children, cloneElement } from 'react'
```
cnairi, line 1: Why do we still need to import React here?

lib/src/components/DatePicker/index.tsx
```diff
@@ -1,18 +1,18 @@
+import type { ReactDatePickerProps } from 'react-datepicker'
+
 import React, { useEffect, useRef, useState } from 'react'
```
cnairi, line 3: We still have the React import here

lib/src/components/DatePicker/index.tsx
```diff
@@ -62,18 +62,18 @@ export const DatePicker = forwardRef<'input', DatePickerProps>(
...
-    const [date, setDate] = useState(formatDate(value))
-    const inputRef = useRef<HTMLInputElement>()
+    const [date, setDate] = useState<Date | null>(formatDate(value))
+    const inputRef = useRef<HTMLInputElement | null>(null)
 
     // format date at component mount
     useEffect(() => {
-      onChange && onChange(formatDate(value))
-      //eslint-disable-next-line
+      onChange?.(formatDate(value))
```
cnairi, line 75: I think it's more readable to have a if condition here (same for all the functions in this file)

lib/src/components/DateTimePicker/index.tsx
```diff
@@ -39,17 +42,17 @@ export const DateTimePicker = forwardRef<'input', DateTimePickerProps>(
     const formatDate: (date: DateTimePickerProps['value']) => ReturnType<typeof getDate> = date =>
       getDate(date, timeIntervals)
 
-    const [date, setDate] = useState(formatDate(value))
+    const [date, setDate] = useState<Date | null>(formatDate(value))
 
     const handleChange: DateTimePickerProps['onChange'] = newDate => {
       setDate(newDate || null)
-      onChange && onChange(newDate && new Date(newDate))
+      onChange?.(newDate && new Date(newDate))
```
cnairi, line 49: I would go with a if here, same as line 54 (and I would keep it consistent across all files)

eslint.config.mjs
```diff
@@ -0,0 +1,64 @@
+import eslint from '@eslint/js'
+import perfectionist from 'eslint-plugin-perfectionist'
+import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended'
+import react from 'eslint-plugin-react'
+import reactHooks from 'eslint-plugin-react-hooks'
+import { globalIgnores } from 'eslint/config'
+import globals from 'globals'
+import tseslint from 'typescript-eslint'
+
+export default tseslint.config(
```
liv7c, line 10: As mentioned during the chapter, it'd be great to add the [eslint-plugin-jsx-a11y](https://www.npmjs.com/package/eslint-plugin-jsx-a11y).

As a start, we could extend the recommended config:

```ts
import jsxA11y from 'eslint-plugin-jsx-a11y';

export default tseslint.config(
	//...
	jsxA11y.flatConfigs.recommended,
)
```

When I tried it out, there were 40 errors (most coming from example files). 

There is also a setting that makes the plugin work better with polymorphic component. Without this setting, you can write stuff like `<Box onClick={}`  and the plugin does not show any warnings.

In our project, we had added that:

```ts
  settings: {
    'jsx-a11y': {
      polymorphicPropName: 'as',
      components: {
        Box: 'div',
        Button: 'button',
        Flex: 'div',
        Grid: 'div',
        Link: 'a',
        Tag: 'div',
        Text: 'p',
      },
    },
  },

```

Available if you want to look to test it out / look at the errors together ☺️

eslint.config.mjs
```diff
@@ -0,0 +1,64 @@
...
+    languageOptions: {
+      ecmaVersion: 'latest',
+      globals: {
+        ...globals.browser,
+        ...globals.node,
+      },
+      sourceType: 'module',
+    },
+    plugins: {
+      react,
```
baptistetruchet, line 27: As mentionned previously, [react/jsx-no-leaked-render](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-leaked-render.md) could be a great addition to our config to prevent leaky `0` or `undefined` displayed in DOM

theo-mesnil: Done

eslint.config.mjs
theo-mesnil: @liv7c i'll add it on another PR on WUI but directly on front library :)

eslint.config.mjs
```diff
@@ -0,0 +1,64 @@
...
+    },
+    rules: {
```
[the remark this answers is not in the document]
theo-mesnil: Yes it will be added to the front config in the same time as @P1X3L scripts for i18n

lib/src/components/DatePicker/index.tsx
```diff
@@ -165,11 +165,12 @@ export const DatePicker = forwardRef<'input', DatePickerProps>(
...
         renderCustomHeader={(props: CustomHeaderProps) => (
           <CustomHeader
-            endYear={endYear}
             isMonthYearPicker={showMonthYearPicker}
             locale={locale}
-            startYear={startYear}
             {...props}
+            // Override props values with our custom ones
+            endYear={endYear}
+            startYear={startYear}
```
guillaumewttj, line 173: This is actually a breaking change 🤔 I have some tests that fails for this reason: https://cloud.cypress.io/projects/9wi68x/runs/3490/overview/ebe04664-a3b1-4529-a457-77a6b6277f95/replay?roarHideRunsWithDiffGroupsAndTags=1&ts=1745338263757.4&att=1

lib/src/components/DateTimePickerCommon/utils.ts
```diff
@@ -1,28 +1,29 @@
...
+  { label: 'December', value: 11 },
 ]
 
 export const DEFAULT_DATE = new Date()
 
-export const getDate = (date: string | number | Date, interval = 15): Date => {
+export const getDate = (date: Date | number | string, interval = 15): Date => {
   if (!date) {
-    return null
+    return new Date()
```
guillaumewttj, line 26: or maybe this one is the breaking change 🤔

lib/src/components/FileDrop/utils.ts
```diff
@@ -8,7 +8,8 @@ export const isAnImage = (file: FileDropChildren['file']): boolean => {
   if (!file) return false
 
   if (typeof file === 'string') {
-    return !!getFileName(file).match(match)
+    const fileName = getFileName(file)
+    return !!fileName && !!fileName.match(match)
```
guillaumewttj, line 12: why do we have that kind of changes ? it is less readable and does not do the same thing as before 🤔

guillaumewttj (commented):
I'm sorry but i'm thinking about doing a revert on that PR, there is too many edits of the logic for a "linters" rework
We need to split that work into parts:
 - The things that are autofixed and the rules behind that*
 - The things that requires human edits + the rules behind that

PR #2742, opened 2025-04-23
fix(AssetDrawer): do not close Drawer when interacting with notifications

SimonDegraeve:
## DESCRIPTION

For the `sourcing` project we have a Talent profile display in an `AssetDrawer` component, in this 
https://linear.app/wttj/issue/SOU-1064/z-index-with-growl-and-modal

## COMPATIBILITY

## QA

- [x] Added tests for all new features


lib/src/components/Drawer/AssetDrawer/index.tsx
```diff
@@ -15,12 +15,24 @@ export const AssetDrawerComponent = forwardRef<'div', AssetDrawerProps>(
...
+        element.contains(target)
+      )
+      return !isTargetWithinPersistentElements
+    }
+
     return (
       <Drawer
         {...rest}
         autoFocusOnShow={false}
+        getPersistentElements={getPersistentElements}
```
anucreative, line 33: We pass a `getPersistentElements` function already from `wttj-front`. Maybe we should put the `{...rest}` at the end of the props to allow overrides ?

anucreative (commented):
Question about usage with other consumers

lib/src/components/Drawer/AssetDrawer/index.tsx
SimonDegraeve: I updated the code so the behavior is the same as prior to this PR when you pass a `getPersistentElements()` prop

lib/src/components/Drawer/AssetDrawer/index.tsx
```diff
@@ -12,15 +12,42 @@ import { CreateWuiProps, forwardRef } from '@/System'
...
+    const getPersistentElements = () =>
+      Array.from(
+        parentGetPersistentElements
+          ? parentGetPersistentElements()
+          : document.querySelectorAll('[data-wui-persistent]')
+      )
+
+    const hideOnInteractOutside =
+      Boolean(parentGetPersistentElements) ||
+      ((event: Event) => {
```
cnairi, line 36: Hi @SimonDegraeve 
I would extract the second condition in another variable for readability like:
```
const hideNotPersistentElement = ((event: Event) => {
      const target = event.target as HTMLElement
        const isTargetWithinPersistentElements = getPersistentElements().some(element =>
          element.contains(target)
        )
        return !isTargetWithinPersistentElements
}
```
(if possible)

SimonDegraeve: Done

PR #2770, opened 2025-06-10
Chore(Docs): update next version to fix vulnerabilities

Isaacmaamouche:
## DESCRIPTION
We had to update our docs' NextJs to at least version 15.2.4 because of vulnerabilities with older version. 
This PR: 
- Bumps Next version
- Updated how our different layout component access params, [next docs](https://nextjs.org/docs/messages/sync-dynamic-apis)
- Fixes [hydration errors with next-themes](https://ui.shadcn.com/docs/dark-mode/next) 

## HOW TO TEST
- poke around
- open the last component you're worked on to ensure it still behaves the same 

## QA
- [x] Dev release

### DIFF
--- a/website/app/components/[id]/layout.tsx
+++ b/website/app/components/[id]/layout.tsx
@@ -6,21 +6,17 @@ import { Text } from '@/Text'
 import { PrevNextPage } from '~/build-app/components/PrevNextPage'
 import { Sidebar } from '~/build-app/components/Sidebar'
 import * as Documentation from '~/build-app/layouts/Documentation'
+import type { Params } from '~/build-app/types'
 import { getPageContent } from '~/build-app/utils/page-content'
 import { getPages } from '~/build-app/utils/pages-components'
 import { getRepository } from '~/build-app/utils/transform-name'
 
 import { Tabs } from './tabs'
 
-type LayoutProps = {
-  children: React.ReactNode
-  params: {
-    id: string
-  }
-}
+type LayoutProps = React.PropsWithChildren<Params>
 
-export async function generateMetadata({ params }: { params: { [key: string]: string } }) {
-  const { id } = params
+export async function generateMetadata({ params }: Params) {
+  const { id } = await params
   const { data } = getPageContent({
     filename: `${getRepository(id)}/docs/index.mdx`,
     isPackage: true,
@@ -34,9 +30,9 @@ export async function generateMetadata({ params }: { params: { [key: string]: st
   }
 }
 
-const Layout = ({ children, params }: LayoutProps) => {
+const Layout = async ({ children, params }: LayoutProps) => {
   const pages = getPages()
-  const { id } = params
+  const { id } = await params
 
   const { data } = getPageContent({
     filename: `${getRepository(id)}/docs/index.mdx`,
--- a/website/app/components/[id]/page.tsx
+++ b/website/app/components/[id]/page.tsx
@@ -2,15 +2,10 @@ import { notFound } from 'next/navigation'
 
 import { Mdx } from '~/build-app/components/Mdx'
 import { TableOfContent } from '~/build-app/components/TableOfContent'
+import type { Params } from '~/build-app/types'
 import { getPageContent } from '~/build-app/utils/page-content'
 import { getPages, getStaticParams } from '~/build-app/utils/pages-components'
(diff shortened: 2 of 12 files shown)


lib/src/components/DateTimePicker/tests/index.test.tsx
```diff
@@ -71,7 +71,7 @@ describe('<DateTimePicker />', () => {
 
     await user.click(datePicker)
 
-    const [monthSelect, yearSelect] = screen.getAllByRole('combobox')
+    const [monthSelect, , yearSelect] = screen.getAllByRole('combobox')
```
anucreative: What's the item we're ignoring when destructuring the elements ?

lib/src/components/Select/index.tsx
```diff
@@ -344,106 +344,108 @@ export const Select = forwardRef<'input', SelectProps>(
...
-                        item: {
-                          label: inputValue,
-                          value: kebabCase(inputValue),
-                        },
-                      })}
-                    >
-                      {renderCreateItem(inputValue)}
-                    </S.Item>
+            <div {...rootProps}>
+              <S.Wrapper {...rootProps} disabled={disabled}>
```
anucreative: This seems "dangerous" to spread `rootProps` on both elements. If I pass `padding=50px` I'll have both elements with padding 50px (so 100px padding in total).

lib/src/components/Select/index.tsx
```diff
@@ -344,106 +344,108 @@ export const Select = forwardRef<'input', SelectProps>(
...
+                <S.InputWrapper>
+                  {isSearchable ? (
+                    <S.Input as="input" type="text" {...inputProps} />
+                  ) : (
+                    <S.Input {...inputProps}>{inputContent}</S.Input>
+                  )}
+                  {icon ? (
```
anucreative: `icon && (` to avoid the ternary ?

lib/src/components/Select/index.tsx
Isaacmaamouche: Oups sorry I fixed that but it got removed when I splitted my branch to isolate the datepicker update 🙈 
It's fixed here: https://github.com/WTTJ/welcome-ui/pull/2771/files#diff-64e9887116fb5c002cbcebf5e492cd9913224081b09f7e9b5526587ba5b38c24R347
but i'll port the change for this PR

lib/src/components/DateTimePicker/tests/index.test.tsx
Isaacmaamouche: I mentionned that in the PR description: 
For some reasons, the combobox is duplicated so we end up with month > month and year > year. The update to the latest version fixes this: https://github.com/WTTJ/welcome-ui/pull/2771)
image
I didn't invest time as the datepicker update fixes this

lib/src/components/Select/index.tsx
Isaacmaamouche: Looks like I can't, we have rule that inforces ternaries: 
image
 @theo-mesnil can you confirm this is intentional ?

theo-mesnil: @anucreative same as jobs-front, it's linter rule, but we can do `!!icon && `

lib/src/components/Select/index.tsx
theo-mesnil: @Isaacmaamouche il faut que dans cette PR il y ai seulement l'upgrade de nextjs et rien d'autre 🙏🏻

lib/src/components/DateTimePicker/tests/index.test.tsx
Isaacmaamouche: I can't consistenly reproduce the issue 🤔 
I'm resetting this change

lib/src/components/Select/index.tsx
Isaacmaamouche: Works fine 🆗 
We have a lot of ternaries, should I fix them all ? In an another PR maybe ?

lib/src/components/Select/index.tsx
Isaacmaamouche: C'était un warning nextjs :/ 
Je fais pas les ternaries non plus du coup ? 
Cleaning this out

theo-mesnil (approved):

PR #2771, opened 2025-06-10
Chore(DatePicker): update datepicker version

Isaacmaamouche:
## DESCRIPTION
This Pr updates the datepicker to its last version. 
Only breaking potential breaking changes I could find was: 
- on [v5.0.0] (https://github.com/Hacker0x01/react-datepicker/releases/tag/v5.0.0) -> Migrate from Popper.js to Floating-UI
- on [v6.6.0] (https://github.com/Hacker0x01/react-datepicker/releases/tag/v6.6.0) -> feat: Replaced classnames with clsx 
I observed no change in behavior for the datepicker 

## HOW TO TEST
- Mess round with the datepicker components 
- Install a dev release to your project and run your tests suites 

## QA
- [ ] Dev release

### DIFF
--- a/lib/package.json
+++ b/lib/package.json
@@ -64,7 +64,7 @@
     "lodash.range": "^3.2.0",
     "lottie-light-react": "^2.4.0",
     "match-sorter": "^8.0.0",
-    "react-datepicker": "^4.16.0",
+    "react-datepicker": "^8.4.0",
     "react-dropzone": "^14.3.5",
     "react-flatten-children": "^1.0.0",
     "react-hot-toast": "^2.4.1"
@@ -78,7 +78,7 @@
     "@types/lodash.debounce": "^4.0.6",
     "@types/lodash.range": "^3.2.9",
     "@types/ramda": "^0.30.2",
-    "@types/react-datepicker": "^4.15.0",
+    "@types/react-datepicker": "^7.0.0",
     "jsdom": "^26.0.0",
     "ramda": "^0.30.1",
     "release-it": "^18.1.2",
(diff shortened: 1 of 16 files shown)


lib/package.json
```diff
@@ -64,7 +64,7 @@
     "lodash.range": "^3.2.0",
     "lottie-light-react": "^2.4.0",
     "match-sorter": "^8.0.0",
-    "react-datepicker": "^4.16.0",
+    "react-datepicker": "^8.4.0",
```
anucreative, line 67: 😱

lib/src/components/DateTimePicker/index.tsx
```diff
@@ -12,10 +13,27 @@ import * as S from './styles'
...
+const getLocale = (locale?: LocalesKeys) => {
+  const browserLocale = navigator.language.split('-')[0]
+
+  if (isValidLocale(locale)) {
+    return locales[locale]
+  }
+  if (isValidLocale(browserLocale)) {
+    return locales[browserLocale]
+  }
+  return locales['enUS']
```
anucreative: It's not `locales['en-US']` ?

anucreative: A test for this would be good. 🤓

anucreative (approved):
👏🏼

lib/src/components/DateTimePicker/index.tsx
Isaacmaamouche: Not according to TS 🤷

Isaacmaamouche: Looks like date-fns uses camel case for locales key: https://github.com/date-fns/date-fns/blob/main/src/locale/en-US/index.ts#L16

PR #2773, opened 2025-06-12
fix: wrong responsiveness on screens

theo-mesnil:
Fix definitions for screens to avoid wrong import loke:

Capture d’écran 2025-06-12 à 15 22 10


anucreative (approved):

P1X3L:
🤯

PR #2790, opened 2025-07-30
Tailwind migration

wttj-deployer:


lib/src/TailwindComponents/ButtonPureCss/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  variant = 'primary',
+  ...rest
+}: ButtonProps<T>) => {
+  const overrides = className.split(' ')
+
+  return (
+    <AriakitButton
+      {...rest}
+      accessibleWhenDisabled={accessibleWhenDisabled}
+      className={cx('root', `variant-${variant}`, `shape-${shape}`, `size-${size}`, ...overrides)}
```
Leiksa: Is there a specific reason to split the classname and then spreading it?
Also there is no need to default the `className` property to an empty string as the utility already handle falsy parameters

lib/src/TailwindComponents/ButtonPureCss/button.module.css
```diff
@@ -0,0 +1,309 @@
+@import 'tailwindcss';
+
+/* The 'components' layer is used so that the styles can be overridden by Tailwind's 'utilities' layer */
+@layer components {
```
Leiksa: Should be renamed by `@layer wui` or something like this to avoid conflict?

lib/src/TailwindComponents/ButtonPureCss/button.module.css
```diff
@@ -0,0 +1,309 @@
...
+
+    /* VARIABLES FROM THEME - DONT CHANGE WITH REACT PROPS */
+    border-width: var(--border-width-sm);
+    transition-property: all;
+    transition-duration: var(--duration-medium);
+    outline: transparent solid var(--border-width-md);
+    font-weight: var(--font-weight-bold);
+    border-radius: var(--radius-md);
+
+    &:focus {
```
Leiksa: This is SCSS specific selector. This is working because the postcss pipeline is still being used or because ultra latest browser start to support it but we definitely shouldn't rely on this and make the file a .SCSS one

lib/src/TailwindComponents/ButtonPureCss/button.module.css
```diff
@@ -0,0 +1,309 @@
...
+      outline-color: var(--outlineColorFocus);
+    }
+
+    &:hover {
+      background-color: var(--color-brand-30);
+      border-color: var(--color-brand-30);
+    }
+
+    &:active {
```
Leiksa:
```suggestion
    &:active:not(:disabled):not([aria-disabled]) {
```

to avoid writing `disabled` styled for active/focus/hover under ⬇️

lib/src/TailwindComponents/ButtonPureCss/button.module.css
Leiksa: (Not linked to your commit Robert)
Also we should use the `&:focus-visible` (to avoid being visible with mouse click)

lib/src/TailwindComponents/ButtonPureCss/button.module.css
```diff
@@ -0,0 +1,309 @@
...
+    transition-duration: var(--duration-medium);
+    outline: transparent solid var(--border-width-md);
+    font-weight: var(--font-weight-bold);
+    border-radius: var(--radius-md);
+
+    &:focus {
+      outline-color: var(--outlineColorFocus);
+    }
+
+    &:hover {
```
Leiksa:
```suggestion
    @media (hover: hover) and (pointer: fine) {
  &:hover {
    blablabla....
  }
}
```

(not linked to your commit Robert) 
We should handle hover state only when it's available (for desktop), to avoid weird behaviour on mobile

lib/src/TailwindComponents/ButtonPureCss/button.module.css
```diff
@@ -0,0 +1,309 @@
...
+      &:focus,
+      &:active {
+        background-color: var(--color-beige-40);
+        border-color: var(--color-beige-40);
+        color: var(--color-beige-70);
+        outline-color: var(--color-beige-10);
+      }
+    }
+
+    &:has(i[data-wui-icon-font]),
```
Leiksa: 💭 Quick question: since :has() is supported by ~93% of browsers, do we have a team policy on using it (progressive enhancement vs. full compatibility)?

lib/src/TailwindComponents/ButtonPureCss/button.module.css
anucreative: Yeah, that might be a good idea. I discovered `@layer` Friday afternoon so I don't have any experience with it. Just have to see how it a layer `wui` would fit with existing layers from Tailwind.

lib/src/TailwindComponents/ButtonPureCss/button.module.css
anucreative: I'd still rather rewrite it as pure css than have SASS code.

lib/src/TailwindComponents/ButtonPureCss/index.tsx
anucreative: Next question is whether we need a utility at all or can we just do string concatenation ?

lib/src/TailwindComponents/ButtonPureCss/button.module.css
```diff
@@ -0,0 +1,309 @@
+@import 'tailwindcss';
```
cnairi: Should we import it here? In my mind, it was more at the level of the consuming app to add this import?

lib/src/TailwindComponents/ButtonPureCss/button.module.css
anucreative: Maybe our rule could / should be based on baseline ? If it's baseline, it's okay to use it ? 

image

lib/src/TailwindComponents/ButtonPureCss/button.module.css
anucreative: In saying that, the sass is much easier to read in cases like this.

anucreative: We can use it in CSS now: 💃🏼 

- https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector
- https://caniuse.com/mdn-css_selectors_nesting

lib/src/TailwindComponents/ButtonPureCss/index.tsx
Leiksa: I don't think my question was related to the utility itself 🤔 
The classNames helper provides type-safe autocompletion for CSS classes and accept conditional `classNames` as we used to do with styledComponent:

```
import styles from './component.module.scss'
const cx = classNames(styles);
//...
<Component className={cx('root', isActive && '--active')} />
// OR
<Component className={cx('root', {
  '--active': isActive
})} />
```

lib/src/TailwindComponents/ButtonPureCss/button.module.css
Leiksa: That's what I said 👀 I agree that native CSS nesting is progressing, but I’d still prefer SCSS for now as 91% supports is not enough to make it widely available.
It’s stable, widely adopted, and offers much more than just nesting (variables, mixins, loops, functions).
For a design system, we need maturity, consistency, and advanced tooling (imo) ✨

lib/src/TailwindComponents/ButtonPureCss/button.module.css
Leiksa: wdyt @cnairi @P1X3L ? 👀

P1X3L: It seems that it has the same compatibility as what we currently also accept regarding our javascript ecosystem, so it looks ok to me.

lib/src/TailwindComponents/ButtonPureCss/index.tsx
cnairi: Hum to answer your question @Leiksa, from what I read, splitting could prevent having duplicated classes?
It could also prevent us from having empty strings/ whitespaces:
const overrides = className.trim().split(/\s+/).filter(Boolean)

Not sure if it's relevant / useful.

What do you think?

Leiksa: @cnairi the utility already handle the duplicated classes & empty string as it stores every classes in a `Set` and then transform it to a string of unique classes -> https://github.com/WTTJ/welcome-ui/blob/chore-tailwind/lib/src/classNames.ts#L20

lib/src/TailwindComponents/ButtonPureCss/button.module.css
P1X3L: Nope this is a mistake since tailwind import has to be made from consuming apps, not our DS. Take a look at how a consuming app handles it here 👉 https://github.com/WTTJ/welcome-ui/blob/caed26cbc6662c75765678f81a316b05c6fe80bd/.storybook/storybook.css

lib/src/TailwindComponents/ButtonPureCss/index.tsx
cnairi: Ok for me, I played a little bit with it to understand better and now I got it 🙏

lib/src/TailwindComponents/Button/index.module.scss
```diff
@@ -0,0 +1,354 @@
...
+    /* VARIABLES THAT CHANGE WITH REACT PROPS */
+    height: var(--height);
+    padding-inline: var(--padding-inline);
+    border-radius: var(--border-radius-md);
+    font-size: var(--font-size);
+    color: var(--color-text-button-primary);
+    background-color: var(--color-background-button-primary-default);
+    border-color: var(--color-border-button-primary-default);
+
+    &:focus,
```
Leiksa:
```suggestion
    &:focus-visible {
      outline-color: var(--color-outline-button-primary-focus);
    }
```
We should use only `focus-visible` because we don't want to have this state when we click but only when we are using a keyboard

lib/src/TailwindComponents/Button/index.module.scss
```diff
@@ -0,0 +1,354 @@
...
+        outline-color: var(--color-outline-button-tertiary-ai-focus);
+      }
+    }
+
+    &-tertiary-danger {
+      background-color: var(--color-background-button-tertiary-danger-default);
+      border-color: var(--color-border-button-tertiary-danger-default);
+      color: var(--color-text-button-tertiary-danger);
+
+      &:active {
```
Leiksa: So, we’re okay with repeating each state across all variants, without relying on variables?

lib/src/TailwindComponents/Button/index.module.scss
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    @media (hover: hover) and (pointer: fine) {
+      &:hover {
+        background-color: var(--color-background-button-primary-hover);
+        border-color: var(--color-border-button-primary-hover);
+      }
+    }
+
+    &:active {
```
Leiksa:
```suggestion
     &:active:not(:disabled):not([aria-disabled]) {
```
To be sure to avoid conflict with disabled state

lib/src/TailwindComponents/ButtonPureCss/button.module.css
cnairi: I resolved this conv as we picked another solution.

lib/src/TailwindComponents/Button/index.module.scss
cnairi: Hey @Leiksa could you resolve here and re-write your comment on: https://github.com/WTTJ/welcome-ui/pull/2808? please? Thanks a lot

lib/src/TailwindComponents/Button/index.module.scss
cnairi: Same here please @Leiksa

lib/src/TailwindComponents/Button/button.module.scss
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  @media (hover: hover) and (pointer: fine) {
+    &:hover {
+      background-color: var(--backgroundColorHover);
+      border-color: var(--borderColorHover);
+    }
+  }
+
+  &:active {
```
Leiksa:
```suggestion
  &:active:not(:disabled):not([aria-disabled]) {
```
To avoid possible conflict with disabled state

lib/src/TailwindComponents/Button/button.module.scss
```diff
@@ -0,0 +1,258 @@
...
+  aspect-ratio: var(--aspectRatio, auto);
+  height: var(--height);
+  padding-inline: var(--paddingInline);
+  border-radius: var(--borderRadius);
+  font-size: var(--fontSize);
+  color: var(--color);
+  background-color: var(--backgroundColor);
+  border-color: var(--borderColor);
+
+  &:focus,
```
Leiksa:
```suggestion
  &:focus-visible {
```
We should use only `focus-visible` because we don't want to have this state when we click but only when we are using a keyboard

stories/Button.stories.tsx
```diff
@@ -0,0 +1,216 @@
+import type { Meta, StoryObj } from '@storybook/react-vite'
```
theo-mesnil: Why it's not on the Button folder ?

cnairi: I cannot remember why during our development process we decided to do so. It was to move on quickly

I just moved it ➡️ [95e6d4b](https://github.com/WTTJ/welcome-ui/pull/2790/commits/95e6d4bb3635f0cf64728437ef40f23839c60ac0)

theo-mesnil: Thaaanks 🔥

lib/src/TailwindComponents/Button/button.module.scss
cnairi: yes [0ce5b05](https://github.com/WTTJ/welcome-ui/pull/2790/commits/0ce5b05111b5995ec0a3f49ed3bb7ca9bd595bac)

lib/src/TailwindComponents/Button/button.module.scss
cnairi: yes [0ce5b05](https://github.com/WTTJ/welcome-ui/pull/2790/commits/0ce5b05111b5995ec0a3f49ed3bb7ca9bd595bac)

lib/src/TailwindComponents/Button/types.ts
```diff
@@ -0,0 +1,21 @@
+import type { ButtonProps as AriakitButtonProps } from '@ariakit/react'
+import type { ButtonHTMLAttributes, CSSProperties, ReactNode } from 'react'
+
+import type {
+  ButtonShapes,
+  ButtonSizes,
+  ButtonVariants,
+  PolymorphicProps,
+} from '../../TailwindTheme/types'
```
theo-mesnil: I think we need to have this types on Button folder not on theme, like we do with styled, and theme theme need to be there also and imported on global theme ?

theo-mesnil: It is logical to note that global CSS does not require buttons variables; they will only be used on this component.

lib/src/classNames.ts
```diff
@@ -0,0 +1,46 @@
+type ClassName<K extends string> = false | K | null | (string & {}) | undefined
```
theo-mesnil: This need to be on utils folder

Leiksa: moved in this PR: https://github.com/WTTJ/welcome-ui/pull/2809

theo-mesnil: Oh yes, the PR that I approved! 😄 🙈

lib/src/TailwindComponents/Button/types.ts
cnairi: Thanks it was fixed by Pix

PR #2792, opened 2025-08-19
chore: proposal to use classnames instead of inline styles

Leiksa:
# Getting closer to CSS standards

## TLDR

This PR aims to:
- Preserve performance by avoiding inline styles created on every render ([React recommendations](https://legacy.reactjs.org/docs/dom-elements.html#style))
- Improve maintainability by keeping components stylistic variants declarative in CSS
- Enhance DX and autocompletion through Typescript types (new classNames utility)

## Issues with the current approach
**1. JS-generated styles -> performance overhead**
In React, each render creates a new `style={...}` object that must be diffed and applied. This is less efficient than static CSS classes.
    - CSS classes are computed and defined once.
    - Inline styles are re-created for every instance of a component. For example, rendering 10 `<Button />` components:
    - With CSS -> styles live in a single stylesheet, reused across all buttons and cacheable if we have the correct infra (not loaded at each page load)
    - With inline styles -> the same style object is recreated 10 times and re-applied on each render, slowing down the view and obviously not cacheable.

**2. Maintainability + specificity**
A design system should ensure scalability and predictable overrides. Inline styles break CSS cascading, complicate debugging, and often require hacks like `!important` to override.

**3. Tokens defined in JS but used as CSS properties**
Tokens are defined in JS, then immediately transformed into CSS custom properties (`var(--token)`). This introduces an unnecessary abstraction layer between the source of truth (CSS variables) and their usage.

## Benefits of the SCCS Modules approach
- Native performance: let the CSS engine handle styles — React doesn't need to inject them at runtime
- Readability & maintainability: stylistic variants are defined declaratively in CS, keeping a clear separation of concerns:
    - Tokens & styles -> handled in CSS/SCSS
    - Behavior & logic -> handled in JS/TS
- Preserved DX: the `classNames` helper provides type-safe autocompletion for CSS classes and accept conditional classNames as we used to do with styledComponent
```tsx
import styles from './component.module.scss'

const cx = classNames(styles);
//...
<Component className={cx('root', isActive && '--active')} />
// OR
<Component className={cx('root', {
  '--active': isActive
})} />

```

- Easier debugging: classNames are visible in the DOM and traceable, unlike opaque inline style objects.

### Before
Screenshot 2025-08-19 at 13 25 33

### After
Screenshot 2025-08-19 at 13 25 45

### DIFF
--- a/lib/src/TailwindComponents/Button/theme.module.css
+++ b/lib/src/TailwindComponents/Button/theme.module.css
@@ -6,8 +6,8 @@
     whitespace-nowrap
     text-ellipsis;
 
-  :global(i[data-wui-icon-font]),
-  :global(svg[data-wui-icon]) {
+  i[data-wui-icon-font],
+  svg[data-wui-icon] {
     display: inline-block;
 
     &:first-child:not(:only-child) {
@@ -34,7 +34,6 @@
     aria-disabled:border-(--borderColorDisabled)
     outline-(length:--border-width-md)
     focus:outline-(--outlineColorFocus)
-    w-(--width)
     h-(--height)
     px-(--paddingInline)
     rounded-(--borderRadius)
--- a/lib/src/TailwindComponents/ButtonFullCss/button.module.scss
+++ b/lib/src/TailwindComponents/ButtonFullCss/button.module.scss
@@ -26,7 +26,6 @@
     disabled:border-(--borderColorDisabled)
     aria-disabled:border-(--borderColorDisabled)
     focus:outline-(--outlineColorFocus)
-    w-(--width)
     h-(--height)
     px-(--paddingInline)
     rounded-(--borderRadius)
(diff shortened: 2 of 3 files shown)


lib/src/TailwindComponents/ButtonProposal/index.tsx
```diff
@@ -0,0 +1,45 @@
+import { Button as AriakitButton } from '@ariakit/react'
+
+import { classNames } from '../../classNames'
+
+import buttonStyles from './button.module.scss'
+import type { ButtonProps } from './types'
+
+const cx = classNames(buttonStyles)
+
+// TODO: use forwardRef from system to bring the polymorphism (as props)
```
cnairi: I wouldn't recommand this option, as forwardRef will be deprecated in a future release: https://react.dev/reference/react/forwardRef

I made a PR here to propose a generic type: [bb2fae0](https://github.com/WTTJ/welcome-ui/pull/2792/commits/bb2fae009e5aeefb009c1a7f4f0a5dbc952efc50)

cnairi (approved):
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

lib/src/TailwindComponents/ButtonProposal/index.tsx
P1X3L: You wanted to highlight only the comment, not the `classNames` part right?

PR #2821, opened 2025-09-15
chore: update aliases config

Leiksa:
## DESCRIPTION

TLDR:
`@*` doesn't work 😎 

**Configuration improvements:**

* Updated the `@` path alias in `tsconfig.json` from `"@*": ["./src/*"]` to `"@/*": ["./src/*"]` for clearer and more accurate path mapping.
* Changed the Vite alias from `'@*': path.resolve(__dirname, './src')` to `'@': path.resolve(__dirname, './src')`, and removed the redundant `'@/utils'` alias, simplifying and standardizing how imports from `src` are handled.


P1X3L (approved):
❤️ 👍

PR #2822, opened 2025-09-15
fix: use inline-flex on button by default

P1X3L:
I initially wanted to handle overflow ellipsis for direct text inside button but it conflicts with the flex based buttons.
So if we want to handle ellipsis on a button, we should use the (yet to come) text component to do so.

### DIFF
--- a/lib/src/components/Button/button.module.scss
+++ b/lib/src/components/Button/button.module.scss
@@ -1,8 +1,8 @@
 .root {
   position: relative;
-  overflow: hidden;
-  white-space: nowrap;
-  text-overflow: ellipsis;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
   cursor: pointer;
   border-style: solid;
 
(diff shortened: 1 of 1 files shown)


lib/src/components/Button/button.module.scss
```diff
@@ -49,16 +49,15 @@
   }
 
   &:has(:global(i.wui-icon-font)),
-  &:has(:global(svg.wui-icon)) {
+  &:has(svg) {
```
P1X3L, line 52: I removed the `.wui-icon` class because am svg as button child might be an icon.
It makes every svg naturally behave correctly 🤷

lib/src/components/Button/button.module.scss
```diff
@@ -49,16 +49,15 @@
...
     /* when button has icon(s) we reduce horizontal padding */
     padding-inline: var(--paddingInlineHasIcon);
   }
 
   /* CHILDREN */
 
   :global(i.wui-icon-font),
-  :global(svg.wui-icon) {
-    display: inline-block;
+  svg {
```
P1X3L, line 60: Same as above

Leiksa (approved):

lib/src/components/Button/button.module.scss
P1X3L: Screenshot 2025-09-15 at 17 40 43

It makes this svg icon (not coming from welcome-ui) work

PR #2830, opened 2025-09-16
chore: migrate Shape component to be tailwind-compatible

anucreative:
## DESCRIPTION

Migrate Shape component to be Tailwind-compatible

Notes:
- Uses inline styles so we can pass a height and width in px

image

## COMPATIBILITY

## QA

### DIFF
--- a/lib/src/components/Shape/index.tsx
+++ b/lib/src/components/Shape/index.tsx
@@ -23,17 +23,15 @@ export const Shape = forwardRef<'div', ShapeProps>(
 
     let style = {}
 
-    if (borderRadius && shape !== 'circle') {
-      style = { ...style, borderRadius: `var(--radius-${borderRadius})` }
-    }
-
     if (maxWidthHeight) {
       style = { ...style, height: maxWidthHeight, width: maxWidthHeight }
     }
 
+    const radius = borderRadius && shape !== 'circle'
+
     return (
       <div
-        className={cx('root', `shape-${shape}`, className)}
+        className={cx('root', `shape-${shape}`, radius && `radius-${borderRadius}`, className)}
         data-testid={dataTestId}
         ref={ref}
         style={style}
--- a/lib/src/components/Shape/index.tsx
+++ b/lib/src/components/Shape/index.tsx
@@ -1,0 +1,47 @@
+import { classNames } from '@/utils'
+import type { CreateWuiProps } from '@old/System'
+import { forwardRef } from '@old/System'
+
+import styles from './shape.module.scss'
+import { getMax } from './utils'
+
+const cx = classNames(styles)
+
+export interface ShapeOptions {
+  borderRadius?: string
+  className?: string
+  h?: string
+  shape?: 'circle' | 'square'
+  w?: string
+}
+
+export type ShapeProps = CreateWuiProps<'div', ShapeOptions>
+
+export const Shape = forwardRef<'div', ShapeProps>(
+  ({ borderRadius, children, className, dataTestId, h, shape = 'square', w }, ref) => {
+    const maxWidthHeight = (w || h) && getMax(w, h)
+
+    let style = {}
+
+    if (borderRadius && shape !== 'circle') {
+      style = { ...style, borderRadius: `var(--radius-${borderRadius})` }
+    }
+
+    if (maxWidthHeight) {
+      style = { ...style, height: maxWidthHeight, width: maxWidthHeight }
+    }
+
+    return (
+      <div
+        className={cx('root', `shape-${shape}`, className)}
+        data-testid={dataTestId}
+        ref={ref}
+        style={style}
+      >
+        {children}
+      </div>
+    )
+  }
+)
+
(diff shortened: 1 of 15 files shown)


lib/src/components/Shape/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { getMax } from './utils'
+
+const cx = classNames(styles)
+
+export interface ShapeOptions {
+  borderRadius?: string
+  className?: string
+  h?: string
+  shape?: 'circle' | 'square'
+  w?: string
```
theo-mesnil, line 15: I think this can be width or height to not have same value as styled

lib/src/components/Shape/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    let style = {}
+
+    if (borderRadius && shape !== 'circle') {
+      style = { ...style, borderRadius: `var(--radius-${borderRadius})` }
+    }
+
+    if (maxWidthHeight) {
+      style = { ...style, height: maxWidthHeight, width: maxWidthHeight }
+    }
```
theo-mesnil: Can this be class ? It's better to avoid inline style

lib/src/components/Shape/shape.module.scss
```diff
@@ -0,0 +1,24 @@
...
+    }
+  }
+
+  .shape-circle {
+    border-radius: 50%;
+  }
+
+  .shape-square {
+    border-radius: 0;
+  }
```
theo-mesnil: Please do same as button example with variable

lib/src/components/Shape/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export interface ShapeOptions {
+  borderRadius?: string
+  className?: string
+  h?: string
+  shape?: 'circle' | 'square'
+  w?: string
+}
+
+export type ShapeProps = CreateWuiProps<'div', ShapeOptions>
```
theo-mesnil, line 18: Is the old way to do, do not use @old project on new components

lib/src/components/Shape/tests/index.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+        {content}
+      </Shape>
+    )
+
+    expect(screen.getByText(content)).toBeVisible()
+  })
+
+  it('should render correctly with borderRadius', () => {
+    render(
+      <Shape borderRadius="lg" dataTestId="shape" h="100px" w="100px">
```
theo-mesnil, line 21:
```suggestion
      <Shape borderRadius="lg" data-testid="shape" h="100px" w="100px">
```

lib/src/components/Shape/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const maxWidthHeight = (w || h) && getMax(w, h)
+
+    let style = {}
+
+    if (borderRadius && shape !== 'circle') {
+      style = { ...style, borderRadius: `var(--radius-${borderRadius})` }
+    }
+
+    if (maxWidthHeight) {
+      style = { ...style, height: maxWidthHeight, width: maxWidthHeight }
```
theo-mesnil, line 27: We need to find a solution to not use style

lib/src/components/Shape/index.tsx
```diff
@@ -0,0 +1,47 @@
+import { classNames } from '@/utils'
+import type { CreateWuiProps } from '@old/System'
+import { forwardRef } from '@old/System'
```
theo-mesnil, line 3:
```suggestion
import { forwardRef } from 'react'
```

website/build-app/examples.js
```diff
@@ -3,18 +3,66 @@
 This file is auto-generate with yarn watch command, do not change it directly!
 **/
 
-import dynamic from "next/dynamic";
+import dynamic from 'next/dynamic'
```
theo-mesnil, line 5: do no save this file, is auto-generate

lib/src/components/Shape/index.tsx
Leiksa: IMO, we could provide a "preset" sizes / arbitrary sizes, that could cover some use cases, but in general the dev could just use tailwind height/width utility classes as `className`

lib/src/components/Shape/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  ({ borderRadius, children, className, dataTestId, h, shape = 'square', w }, ref) => {
+    const maxWidthHeight = (w || h) && getMax(w, h)
+
+    let style = {}
+
+    if (maxWidthHeight) {
+      style = { ...style, height: maxWidthHeight, width: maxWidthHeight }
+    }
+
+    const radius = borderRadius && shape !== 'circle'
```
Leiksa, line 30: Same here for the border-radius, we could just use in the implementation the [tailwind radius](https://tailwindcss.com/docs/border-radius) utility classes

anucreative:
Shape will be deprecated

PR #2834, opened 2025-09-17
chore: migrate input text to tailwind

jeromeraffin:
This pull request introduces a new `InputText` component to the codebase, complete with implementation, documentation, examples, styles, type definitions, and tests. It also integrates the component into the website's global styles and example loader, ensuring it is available for documentation and demonstration purposes.

**New InputText Component Implementation and Integration**

* Added the `InputText` component in `lib/src/components/InputText/index.tsx`, supporting custom variant and sizes. And with disabled, focused and hover states.
* Defined prop types and options for `InputText` in `lib/src/components/InputText/types.ts`, including support for custom size and variant.
* Created comprehensive SCSS styles for the `InputText` component, including size variants and icon positioning, in `lib/src/components/InputText/input-text.module.scss`.
* Added a test to verify basic rendering of the `InputText` component in `lib/src/components/InputText/tests/index.test.tsx`.

**Documentation and Examples**

* Authored a detailed MDX documentation page for `InputText` in `lib/src/components/InputText/docs/index.mdx`, describing usage, variants, and properties.
* Documented the `InputText` component's properties in JSON format for automated docs in `lib/src/components/InputText/docs/properties.json`.

**Website Integration**

* Registered all new `InputText` example files in the website's example loader (`website/build-app/examples.js`) for live documentation.
* Included `input-text.module.scss` in the global CSS for the website to ensure consistent styling.

**Other Minor Changes**

* Removed all input html default attributes from the internal props options (data-testid, onBlur, autoFocus ...)
* The documentation contains examples of the use of `InputText` only, the other use cases should be defined in the `Field` component
* Create a `default` variant to be able to override border color and outline color with other variants

### DIFF
--- a/lib/src/components/InputText/index.tsx
+++ b/lib/src/components/InputText/index.tsx
@@ -1,0 +1,80 @@
+import React, { forwardRef } from 'react'
+
+import { classNames } from '@/utils'
+// @WUI-214 - replace old CleanButton component when it will be updated
+import { ClearButton } from '@old/ClearButton'
+
+import inputTextStyles from './input-text.module.scss'
+import type { InputTextProps, Size } from './types'
+
+const cx = classNames(inputTextStyles)
+
+const FIELD_ICON_SIZE: { [key in Size]: string } = { lg: 'sm', md: 'sm', sm: 'sm', xs: 'xs' }
+
+export const InputText = forwardRef<HTMLInputElement, InputTextProps>(
+  (
+    {
+      className,
+      icon,
+      iconPlacement = 'left',
+      isClearable,
+      name,
+      onChange,
+      size = 'md',
+      transparent,
+      type = 'text',
+      value,
+      variant = 'default',
+      ...rest
+    },
+    ref
+  ) => {
+    const hasIcon = icon && iconPlacement
+    const hasRightIcon = hasIcon && iconPlacement === 'right'
+    const hasLeftIcon = hasIcon && iconPlacement === 'left'
+    const iconSize = FIELD_ICON_SIZE[size]
+
+    const handleReset = () => {
+      const event = {
+        preventDefault: () => null,
+        target: { name, value: '' },
+      } as React.ChangeEvent<HTMLInputElement>
+      if (onChange) onChange(event)
+    }
+
+    return (
+      <div className={cx('input-text-wrapper')}>
+        <input
+          {...rest}
+          className={cx(
+            'root',
+            `variant-${variant}`,
+            `size-${size}`,
+            transparent && 'transparent',
+            hasIcon && `placement-${iconPlacement}`,
+            className
+          )}
+          id={name}
+          name={name}
+          onChange={onChange}
+          ref={ref}
+          type={type}
+          value={value}
+        />
+
+        {hasLeftIcon ? (
+          <div className={cx('icon-wrapper', `icon-placement-left-${iconSize}`)}>
+            {React.cloneElement(icon, { ...icon.props, size: iconSize })}
+          </div>
+        ) : null}
+
+        {isClearable || hasRightIcon ? (
+          <div className={cx('icon-wrapper', `icon-placement-right-${iconSize}`)}>
+            {isClearable && value ? <ClearButton onClick={handleReset} /> : null}
+            {hasRightIcon ? React.cloneElement(icon, { ...icon.props, size: iconSize }) : null}
+          </div>
+        ) : null}
+      </div>
+    )
+  }
+)
--- a/lib/src/components/InputText/input-text.module.scss
+++ b/lib/src/components/InputText/input-text.module.scss
@@ -1,0 +1,184 @@
+.input-text-wrapper {
+  position: relative;
+}
+
+.root {
+  border-radius: var(--radius-md);
+  border-style: solid;
+  border-width: var(--border-width-sm);
+  font-size: var(--font-size-sm);
+  font-weight: var(--font-weight-regular);
+  line-height: var(--line-height-16);
+  outline: transparent solid var(--border-width-md);
+  width: 100%;
+  transition: var(--duration-medium);
+  appearance: none;
+
+  /* VARIABLES THAT CHANGE WITH REACT PROPS */
+  background-color: var(--backgroundColor);
+  color: var(--color);
+  border-color: var(--borderColor);
+  height: var(--height);
+  padding-inline: var(--paddingInline);
+  padding-block: var(--paddingBlock);
+
+  --borderColor: var(--color-neutral-30);
+  --backgroundColor: var(--color-neutral-10);
+  --color: var(--color-neutral-90);
+
+  &:disabled,
+  &[aria-disabled] {
+    --backgroundColor: var(--color-beige-40);
+    --color: var(--color-beige-70);
+    cursor: not-allowed;
+  }
+
+  &:hover {
+    --borderColor: var(--borderColorHover);
+  }
+
+  &[data-focus-visible],
+  &:focus-visible {
+    --borderColor: var(--borderColorFocused);
+    outline-color: var(--outlineColorFocus);
+  }
+
+  &::placeholder {
+    --color: var(--color-neutral-60);
+  }
+}
+
+.variant {
+  &-default {
+    --borderColorHover: var(--color-neutral-40);
+    --outlineColorFocus: var(--color-brand-20);
+    --borderColorFocused: var(--color-brand-40);
+  }
+
+  &-danger {
+    --borderColor: var(--color-red-70);
+    --borderColorHover: var(--color-red-70);
+    --borderColorFocused: var(--color-red-70);
+    --outlineColorFocus: var(--color-red-30);
+  }
+
+  &-warning {
+    --borderColor: var(--color-orange-60);
+    --borderColorHover: var(--color-orange-60);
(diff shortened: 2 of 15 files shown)


lib/src/components/InputText/docs/examples/hint.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+const Example = () => {
+  return (
+    <Field hint="Your hint">
+      <InputText placeholder="Welcome" />
+    </Field>
+  )
+}
+
+export default Example
```
cnairi: I think that more than a "Hint" example, we want a "Compound with Field" example.

Like a full example with a label + a hint, and even maybe with the possible Field variant.

We can also decide to keep it simple and only demonstrate the full variants in the field doc.

But I would at least make the hint example more generic to show that you can compound with the Field props

lib/src/components/InputText/index.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { forwardRef } from 'react'
+
+import { classNames } from '@/utils'
+import { ClearButton } from '@old/ClearButton'
```
cnairi, line 5: Should we add a todo to clean this one?

lib/src/components/InputText/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+import inputTextStyles from './input-text.module.scss'
+import type { InputTextProps, Size } from './types'
+
+const cx = classNames(inputTextStyles)
+
+export const FIELD_ICON_SIZE: { [key in Size]: string } = { lg: 'sm', md: 'sm', sm: 'sm', xs: 'xs' }
```
cnairi: Why not `export const FIELD_ICON_SIZE: { [key in Size]: Size } = { lg: 'sm', md: 'sm', sm: 'sm', xs: 'xs' }`?

lib/src/components/InputText/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+              ? React.cloneElement(icon, { ...icon.props, size: iconSize })
+              : null}
+          </div>
+        ) : null}
+      </div>
+    )
+  }
+)
+
+InputText.displayName = 'InputText'
```
cnairi: Not sure the displayName is still necessary

lib/src/components/InputText/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const handleReset = () => {
+      const event = {
+        preventDefault: () => null,
+        target: { name, value: '' },
+      } as React.ChangeEvent<HTMLInputElement>
+      if (onChange) onChange(event)
+    }
+
+    return (
+      <div className="input-text-wrapper">
```
cnairi: We don't need to use the cx here to bind to our css module?
`<div className={cx("input-text-wrapper")}>`

lib/src/components/InputText/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <input
+          {...rest}
+          className={cx(
+            'root',
+            variant && `variant-${variant}`,
+            transparent && 'transparent',
+            size && `size-${size}`,
```
cnairi: You have a default variant and a default size here, so you don't need the conditions
```
className={cx(
            'root',
            `variant-${variant}`,
            `size-${size}`,
             transparent && 'transparent',
```

lib/src/components/InputText/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+        {hasIcon && iconPlacement === 'left' ? (
+          <div className={cx('icon-wrapper', `icon-placement-left-${iconSize}`)}>
+            {React.cloneElement(icon, { ...icon.props, size: iconSize })}
+          </div>
+        ) : null}
+
+        {isClearable || (hasIcon && iconPlacement === 'right') ? (
+          <div className={cx('icon-wrapper', `icon-placement-right-${iconSize}`)}>
+            {isClearable && value ? <ClearButton onClick={handleReset} /> : null}
+            {hasIcon && iconPlacement === 'right'
```
cnairi: I think you could extract this condition in a variable, as it's used twice and also for more readability with the nested ternary:
`const hasIconRight = hasIcon && iconPlacement === 'right'`

lib/src/components/InputText/input-text.module.scss
```diff
@@ -0,0 +1,183 @@
...
+  position: relative;
+}
+
+.root {
+  border-radius: var(--radius-md);
+  border-style: solid;
+  border-width: var(--border-width-sm);
+  font-size: var(--font-size-sm);
+  font-weight: var(--font-weight-regular);
+  line-height: 1rem;
```
cnairi: This corresponds to the theme value `--line-height-16`

lib/src/components/InputText/input-text.module.scss
```diff
@@ -0,0 +1,183 @@
...
+  outline: transparent solid var(--border-width-md);
+  width: 100%;
+  transition: medium;
```
cnairi: Shouldn't it be `var(--duration-medium)` instead?

lib/src/components/InputText/input-text.module.scss
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  &-right.size {
+    &-xs {
+      padding-right: 1.75rem;
+    }
+    &-sm,
+    &-md,
+    &-lg {
+      padding-right: 2.25rem;
```
cnairi, line 99: Should all these be added to theme values?
--spacing-28: 1.75rem; /* 28px */
--spacing-36: 2.25rem; /* 36px */

lib/src/components/InputText/tests/index.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const input = screen.getByTestId('input') as HTMLInputElement
+    expect(input.value).toBe('test')
+
+    const clearButton = screen.getByRole('button')
+
+    await user.click(clearButton)
+
+    expect(input.value).toBe('')
+  })
+})
```
cnairi, line 52: I think we could add 3 tests here:
- One on the variant property
- One on the disabled property
- One on the sizes property

This is non blocker

lib/src/components/InputText/types.ts
```diff
@@ -0,0 +1,15 @@
+export interface InputTextOptions {
+  icon?: JSX.Element
+  iconPlacement?: 'left' | 'right'
```
Leiksa, line 3: I'm wondering if we should update the API here. Instead of scoping it strictly as an icon, maybe we could have `right` and `left` props that accept a `JSX.Element`. In some cases, we’re actually using a `button` or a `tooltip`, so it feels a bit confusing.
Also, having two separate props to handle both placement and a JSX.Element doesn’t seem very consistent 🤔
WDYT @cnairi @jeromeraffin ?

lib/src/components/InputText/tests/index.test.tsx
jeromeraffin: @cnairi I think that we should have a more global strategy on this kind of tests, @P1X3L had some troubles on testing if a color was correctly set on a component

lib/src/components/InputText/index.tsx
jeromeraffin: no need to export it, you're right

lib/src/components/InputText/index.tsx
jeromeraffin: Will add a comment on this one

lib/src/components/InputText/docs/examples/hint.tsx
jeromeraffin: Will remove this example, it doesn't make sense with the topic we commented on slack

lib/src/components/InputText/types.ts
jeromeraffin: totally agree with this, do you think we can validate this PR and make a breaking change PR after?

lib/src/components/InputText/index.tsx
cnairi: I was talking about this part (replace string with Size which is more accurate) `[key in Size]: Size`

lib/src/components/InputText/tests/index.test.tsx
cnairi: Oki we'll see later, should we create a subticket then?

lib/src/components/InputText/types.ts
cnairi: What's the main idea?
Like using a compound pattern? With LeftElement RightElement?

And yes for doing it later

lib/src/components/InputText/input-text.module.scss
cnairi: You can also do this one later @jeromeraffin no emergency, you can merge it

lib/src/components/InputText/types.ts
Leiksa: Yes, exactly 😄 The idea is to simplify the API by removing the strict “icon” scope. 
Instead of having two props (icon + iconPlacement) to handle both the JSX element and its placement, we could have just `left` and `right` (or whatever other name) props that accept any `JSX.Element`. That way, we can directly pass buttons, tooltips, or icons without forcing them into an “icon” prop, and they’ll automatically be placed correctly
This is a breaking change, so we can handle it later of course, but I think it’s also important to take the time to make our component more consistent 👀

PR #2837, opened 2025-09-17
feat: add svg icons

theo-mesnil:
- Remove IconFont script & circle
- Rework script to optimize and create icons
- Change icons on IconVariant & Button / Website


scripts/icons-collect.mjs
```diff
@@ -54,12 +54,13 @@ const writeIconIndex = (outputFolder, iconName) => {
...
+import { Icon } from '../../Icon'
+import type { IconProps } from '../../Icon.types'
 
 import content from './content.json'
 
 export const ${iconName}Icon: React.FC<IconProps> = props => {
-  return <Icon data-wui-icon alt="${iconName}" content={content} {...props} />
+  return <Icon alt="${iconName}" content={content} {...props} />
 }
 `
```
cnairi, line 63: Really non blocker but for the button we used a different syntax for typing the component:
```
import { Icon } from '../../Icon'
import type { IconProps } from '../../Icon.types'

import content from './content.json'

export const ActionsIcon = (props: IconProps) => {
  return <Icon alt="Actions" content={content} {...props} />
}
```

cnairi (approved):

lib/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,32 @@
+import { forwardRef } from 'react'
+
+import { classNames } from '@/utils'
+
+import iconStyles from './icon.module.scss'
```
P1X3L, line 5: where are the iconStyles? I don't see the related icon.module.scss!

lib/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,32 @@
...
+import type { IconProps } from './Icon.types'
+
+const cx = classNames(iconStyles)
+
+export const Icon = forwardRef<SVGSVGElement, IconProps>(
```
P1X3L, line 10: I think you can safely remove `forwardRef` given that it is deprecated with react 19.

lib/src/components/Icon/Icon.tsx
cnairi: ➡️ `welcome-ui/lib/src/components/Icon/icon.module.scss`

lib/src/components/Icon/Icon.tsx
cnairi: In this slack channel: https://wttj.slack.com/archives/C09CT2BSF0B/p1758015552328139

We decided to remain compatible with react 18

P1X3L: Ok sorry, I missed it

lib/src/components/Icon/Icon.tsx
P1X3L: Ok sorry 🙈

PR #2850, opened 2025-09-18
feat: migrate Label

theo-mesnil:


lib/src/components/Icon/Icon.test.tsx
```diff
@@ -1,5 +1,6 @@
-import { render, screen } from '@testing-library/react'
-import { describe, expect, it } from 'vitest'
+import { screen } from '@testing-library/react'
+
+import { render } from '../../../tests'
```
cnairi: import { render } from '@tests'

lib/src/components/Label/index.test.tsx
```diff
@@ -0,0 +1,39 @@
+import { screen } from '@testing-library/react'
+
+import { render } from '../../../tests'
```
cnairi: import { render } from '@tests'

lib/src/components/Label/types.ts
```diff
@@ -0,0 +1,13 @@
+import type { ComponentPropsWithRef } from 'react'
+
+export type LabelProps = ComponentPropsWithRef<'label'> &
+  LabelOptions & {
+    className?: string
+  }
```
cnairi: className is already defined in ReactHTMLAttribute (or some react generic type)

cnairi (approved):

PR #2858, opened 2025-09-19
chore: migrate Tabs

anucreative:
## DESCRIPTION

No change to API

image

## COMPATIBILITY

## QA


lib/src/components/Tabs/components/ActiveBar/index.tsx
```diff
@@ -0,0 +1,77 @@
+import React, { useState } from 'react'
+
+import { classNames } from '@/utils'
+import { useIsomorphicLayoutEffect, useViewportSize } from '@old/utils'
```
cnairi: Hey, we are not supposed to import old things from our new components.

You need to migrate these parts as well

lib/src/components/Tabs/components/ActiveBar/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  activeTab: HTMLElement
+  listRef: React.MutableRefObject<undefined>
+  orientation: Omit<UseTabState['orientation'], 'both'>
+}
+
+export interface ActiveBarReturn {
+  offset?: number
+  orientation?: Omit<UseTabState['orientation'], 'both'>
+  size?: number
+}
```
cnairi: Could you move those types to a types.ts file please?

lib/src/components/Tabs/components/ActiveBar/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+      transform: `translateY(${activeBar.offset}px)`,
+    }
+  } else {
+    style = {
+      transform: `translateX(${activeBar.offset}px)`,
+      width: activeBar.size,
+    }
+  }
+
+  return <span className={cx('root')} style={style} />
```
cnairi:
```  
let style = {}
  if (orientation === 'vertical') {
    style = {
      '--height': `${activeBar.size}px`,
      '--transform': `translateY(${activeBar.offset}px)`,
    }
  } else {
    style = {
      '--transform': `translateX(${activeBar.offset}px)`,
      '--width': `${activeBar.size}px`,
    }
  }

  return <span className={cx('root', `orientation-${orientation}`)} style={style} />
```

And then in your scss file:
```
    /* Properties compiled at run time, based on the real tab size */
    height: var(--height, auto);
    width: var(--width, auto);
    transform: var(--transform, auto);
```

lib/src/components/Tabs/components/TabList/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+import styles from './tab-list.module.scss'
+
+export type Size = 'md' | 'sm'
+
+export type TabListProps = {
+  children: React.ReactNode
+  className?: string
+  size?: Size
+  store: UseTab
+}
```
cnairi: Can you put all the types elated to tabs in their own types.ts file please?

lib/src/components/Tabs/components/TabList/index.tsx
```diff
@@ -0,0 +1,49 @@
+import * as Ariakit from '@ariakit/react'
```
cnairi: As you only need two elements from ariakit, you could replace it with 
`import { TabList as AriakitTabList, useStoreState } from '@ariakit/react'`

lib/src/components/Tabs/components/TabList/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const listForkedRef = useForkRef(ref, listRef)
+
+    const { orientation, selectedId } = Ariakit.useStoreState(store)
+    const { activeTab, tabs } = useTrackActiveTabs(selectedId, children)
+
+    const hasMultipleTabs = React.Children.count(children) > 1
+
+    return (
+      <Ariakit.TabList
+        className={cx('root', size && `size-${size}`, className)}
```
cnairi: You have a default size, so you could just `className={cx('root', `size-${size}`, className)}`

lib/src/components/Tabs/components/TabList/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+        {tabs}
+        {hasMultipleTabs ? (
+          <ActiveBar activeTab={activeTab} listRef={listRef} orientation={orientation} />
+        ) : null}
+      </Ariakit.TabList>
+    )
+  }
+)
+
+TabList.displayName = 'TabList'
```
cnairi: I'm not sure the displayName is necessary anymore

lib/src/components/Tabs/components/TabList/tab-list.module.scss
```diff
@@ -0,0 +1,21 @@
+@layer components {
+  .root {
+    width: 100%;
+    position: relative;
+    display: flex;
+    overflow-x: auto;
+    border: 0;
+    gap: var(--spacing-xl);
+    box-shadow: inset 0 calc(-1 * var(--border-width-sm)) 0 var(--color-neutral-30);
+    font-size: var(--font-size-md);
```
cnairi: box-shadow: var(--boxShadow, inset 0 calc(-1 * var(--border-width-sm)) 0 var(--color-neutral-30))
font-size: var(--fontSize, --font-size-md)

and then in line 14 `--boxShadow: inset calc(-1 * var(--border-width-sm)) 0 0 var(--color-neutral-30);``

Line 19 `--fontSize: var(--font-size-sm);`

lib/src/components/Tabs/components/TabPanel/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import type { UseTab } from '../..'
+
+import styles from './tab-panel.module.scss'
+
+export type TabPanelProps = {
+  children: React.ReactNode
+  store: UseTab
+  tabId?: string
+}
```
cnairi: Same for this type

lib/src/components/Tabs/components/TabPanel/index.tsx
```diff
@@ -0,0 +1,28 @@
+import * as Ariakit from '@ariakit/react'
```
cnairi: `import { TabPanel as AriakitTabPanel } from '@ariakit/react'`

lib/src/components/Tabs/docs/examples/badge.tsx
```diff
@@ -0,0 +1,46 @@
+import { Tab, useTab } from '@/components/Tabs'
+import { Badge } from '@old/Badge'
+import { HeartIcon } from '@old/Icons'
```
cnairi: Icons are already migrated `import { HeartIcon } from '@/components/Icons'`

lib/src/components/Tabs/docs/examples/vertical.tsx
```diff
@@ -0,0 +1,39 @@
+import { Tab, useTab } from '@/components/Tabs'
+import { Box } from '@old/Box'
```
cnairi: We shouldn't import old components anymore if they are already migrated or if they are going to disappear

You can just use `<div className="flex">` instead

lib/src/components/Tabs/index.tsx
```diff
@@ -0,0 +1,56 @@
+import * as Ariakit from '@ariakit/react'
```
cnairi:
```
import { Tab as AriakitTab } from '@ariakit/react'
import type {
  TabStoreProps as AriakitTabStoreProps,
  TabStore,
  TabStoreProps,
  TabStoreState,
} from '@ariakit/react'
```

lib/src/components/Tabs/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type Size = 'md' | 'sm'
+
+interface TabOptions {
+  children: React.ReactNode
+  className?: string
+  id?: string
+  size?: Size
+  store: UseTab
+}
```
cnairi: all the types here should go in a types.ts file please

lib/src/components/Tabs/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const cx = classNames(styles)
+
+export const TabComponent = React.forwardRef(
+  <T extends React.ElementType = 'button'>(
+    { as, children, className, id, store, ...rest }: TabProps<T> & { as?: T },
+    ref: React.Ref<HTMLElement>
+  ) => {
```
cnairi: Diane created a utils: forwardRefWithAs to handle this case

lib/src/components/Tabs/theme.ts
```diff
@@ -0,0 +1,101 @@
...
+    tabsBorder: {
+      horizontal: {
+        boxShadow: `inset 0 -${borderWidths.sm} 0 ${colors['neutral-30']}`,
+      },
+      vertical: {
+        boxShadow: `inset -${borderWidths.sm} 0 0 ${colors['neutral-30']}`,
+      },
+    },
+  }
+}
```
cnairi: This file should be removed

lib/src/components/Tabs/components/ActiveBar/index.tsx
anucreative: How is this better ?

lib/src/components/Tabs/components/TabList/tab-list.module.scss
anucreative: 🤔

lib/src/components/Tabs/components/ActiveBar/active-bar.module.scss
```diff
@@ -0,0 +1,23 @@
...
+    position: absolute;
+    will-change: width, transform;
+    transition-property: all;
+    transition-duration: var(--duration-medium);
+    transition-timing-function: var(--timing-primary);
+    background-color: var(--color-brand-40);
+
+    [aria-orientation='horizontal'] & {
+      height: var(--border-width-md);
+
```
Isaacmaamouche: empty line 🙈

lib/src/components/Tabs/components/ActiveBar/index.tsx
Isaacmaamouche: What if if set two variables in style 
```typescript 
style={{
--offset: activebar.offset
--size: activeBar.size
}}
```
and in css, using .orientation-XXX in the .scss file, 
we apply either height & translateY or width & translateX 


```css
&.orientation-vertical {
     height: var( --size),
     transform: translateY(var(--offset)),
    }

&.orientation-horizontal  {
     width: var( --size),
     transform: translateX(var(--offset)),
    }
```

lib/src/components/Tabs/components/TabList/tab-list.module.scss
Isaacmaamouche: I think shee means that we should declare `box-shadow` and `font-size` once in .root with a var as a value, then use the classes line 12 and 18 to reassign the var value. 

I see that only the bow shadow direction changes, that tricky to read in itself, so I agree, it's not easier on the eyes

```css

.root {

/* base value in variable */
--root-box-shadow-direction: inset 0 calc(-1 * var(--box-shadow-width)) 0 var(--color-neutral-30);
box-shadow: var(--root-box-shadow-direction);

/* base value in variable */
--root-font-size: var(--font-size-md;
 font-size: var(--root-font-size);

&[aria-orientation='vertical'] {
/* variable updated because of the orientation modifier */
--root-box-shadow-direction: inset calc(-1 * var(--border-width-sm)) 0 0 var(--color-neutral-30);
  }


.size-sm {
/* variable updated because of the orientation modifier */
--root-font-size: var(--font-size-sm);
}

}
```

anucreative: I still don't get this. We're saying that _every_ css property that _can be overridden_  needs a css variable ? Even if there are only 2 possible options (vertical and horizontal, md and sm) ?

Isaacmaamouche:
🔥

lib/src/components/Tabs/components/ActiveBar/index.tsx
anucreative: The Tailwind docs use this exact example: https://tailwindcss.com/docs/styling-with-utility-classes#when-to-use-inline-styles
```
      style={{
        backgroundColor: buttonColor,
        color: textColor,
      }}
```

My problem is that by setting a variable in one file and using it in another (without passing it or importing it) breaks any link / relationship between them. This means it's very easy to set variables and not use them (redundant) or forgetting to set them so they're not available in the CSS.

lib/src/components/Tabs/components/TabList/tab-list.module.scss
Isaacmaamouche: I'd say yes -> modifiers = modifiers "pattern" - at least for `font-size` as `size` has multiple values. 
For the `box-shadow`, it's either vertical or horizontal -> What if we don't set anything in .root, and declare values for both states in their own selector ? No silly one-time use css vars

lib/src/components/Tabs/types.ts
```diff
@@ -0,0 +1,44 @@
...
+  store: TabStore
+}
+
+export type TabPanelProps = {
+  children: React.ReactNode
+  store: UseTab
+  tabId?: string
+}
+
+export type TabProps<T extends React.ElementType = 'button'> = PolymorphicProps<T> &
```
Leiksa, line 42: We don't need `PolymorphicProps<T>` as the `forwardRefWithAs` handle the polymorphism

lib/src/components/Tabs/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  //If the scale property is not invalid or undefined, return its value
+  //Else check the parent's scale property
+  if (!isNaN(elementScale)) {
+    return elementScale
+  } else {
+    return getParentScale(element.parentElement)
+  }
+}
+
+export function useForkRef(
```
Leiksa, line 21: Why this utils is in the `Tabs` scope? We should put it in the common

lib/src/components/Tabs/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export function useTrackActiveTabs(
+  selectedId: UseTabState['selectedId'],
+  children: React.ReactNode
+): { activeTab: HTMLElement | null; tabs: React.ReactNode } {
+  const [activeTab, setActiveTab] = useState<HTMLElement | null>(null)
+
+  const tabs = React.Children.map(children, (child: React.ReactElement) => {
+    if (child.props.id === selectedId) {
+      return cloneElement(child, { ref: setActiveTab })
```
Leiksa, line 49: We need to merge the child `ref` with the setActive otherwise we are losing it

lib/src/components/Tabs/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+    return child
+  })
+
+  return { activeTab, tabs }
+}
+
+function setRef(ref: ForwardedRef<unknown>, value: unknown): void {
```
Leiksa, line 57: Should be in the common util too?

PR #2859, opened 2025-09-19
Chore: migrate select component to tailwind [WUI-171]

Isaacmaamouche:
This pull request migrates the `Select` component, covering its main features, customization options, example files demonstrating different configurations.

CHANGES: 
- Icons use to accept `title`
I replace all instances`alt`
they turn into an `aria-label`

TODO
- [X] How to type the input without omitting homonym props: [Select/types.ts](https://github.com/WTTJ/welcome-ui/blob/WUI-171-chore-migrate-select-tailwind/lib/src/components/Select/types.Ts#L58-L59)
-> used Diane's `MergeProps` utils
- [X] How to type the forwardRef to avoid casting [Select/index.tsx](https://github.com/WTTJ/welcome-ui/blob/WUI-171-chore-migrate-select-tailwind/lib/src/components/Select/index.tsx#L100-L101)
-> I kept the type definition/casting from the old select `ref: React.MutableRefObject<HTMLInputElement>`
- [x] How & where do we pass classNames ?


lib/src/components/Select/index.tsx
```diff
@@ -0,0 +1,426 @@
...
+      renderGroupHeader,
+      renderItem = itemToString,
+      renderMultiple = multipleSelections,
+      size = 'md',
+      transparent,
+      value: defaultSelected,
+      variant,
+      ...rest
+    }: SelectProps,
+    ref: React.MutableRefObject<HTMLInputElement>
```
Leiksa: Why did you type the `ref`? 🤔

lib/src/components/Select/index.tsx
```diff
@@ -0,0 +1,426 @@
...
+  ): JSX.Element => {
```
Leiksa: Same as above here? 🤔

lib/src/components/Select/index.tsx
```diff
@@ -0,0 +1,426 @@
...
+        if (isSearchable) setIsOpen(true)
+      }
+    }, [isSearchable, autoFocus, ref])
+
+    // Ensure values are controlled by parent
+    useEffect(() => {
+      setSelected(defaultSelecteds)
+      setInputValue(defaultInputValue)
+      setOptions(defaultOptions)
+    }, [defaultInputValue, defaultOptions, defaultSelecteds])
```
Leiksa: Why do we name them as default values if they are reactive?

lib/src/components/Select/index.tsx
```diff
@@ -0,0 +1,426 @@
...
+            ref,
+            size,
+            tabIndex: 0,
+            transparent,
+            variant: isOpen ? 'focused' : variant,
+            ...rest,
+          }) as React.InputHTMLAttributes<HTMLInputElement>
+          const iconSize = FIELD_ICON_SIZE[size]
+
+          const inputClassnames = cx(
```
Leiksa: This should be outside of the render function as we don't use any parameter

Leiksa: Maybe every props that are not related to the render function parameters should be outside the render function

lib/src/components/Select/index.tsx
```diff
@@ -0,0 +1,426 @@
...
+                ) : null}
+                <div className={cx('indicators', `size-${size}`)}>
+                  {isShowDeleteIcon ? DeleteIcon : null}
+                  {ArrowIcon}
+                </div>
+              </div>
+              {isShowMenu ? (
+                <ul className={cx('menu')} {...getMenuProps()}>
+                  {
+                    options.reduce(
```
Leiksa: If there are not much parameters usage from the render function we could create a `useCallback` outside the render function named `renderOptions` that take the `getItemProps` function as parameter and use it here. In this way we will not recreate every reference for each re-render

lib/src/components/Select/types.ts
```diff
@@ -0,0 +1,68 @@
...
+  hasIcon?: boolean
+  icon?: React.ReactElement
+  iconPlacement?: 'both' | 'left' | 'right'
+  id?: string
+  isClearable?: boolean
+  isCreatable?: boolean
+  isMultiple?: boolean
+  isSearchable?: boolean
+  name?: string
+  onBlur?: () => void
```
Leiksa: `onBlur`, `onClick`, `onFocus`, `placeholder`, `name`, `disabled`, `id`... should already be in the `ComponentProps<input>`

lib/src/components/Select/docs/examples/allow-unselect-from-list.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const handleChange = (newValue: SelectProps['value']) => {
+    setValue(newValue)
+  }
+
+  return (
+    <Select
+      allowUnselectFromList
+      disableCloseOnSelect
+      icon={<LinkIcon alt="Social networks" color="neutral-90" />}
```
Leiksa, line 27: `Icon` should not have `alt` attribute as they are not `img` but simple element. Also in this context they are simply decorative so we don't need to put any attribute that give context to the screen reader

lib/src/components/Select/docs/examples/allow-unselect-from-list.tsx
```diff
@@ -0,0 +1,46 @@
...
+      isMultiple
+      isSearchable
+      name="welcome"
+      onChange={handleChange}
+      options={ITEMS}
+      renderItem={(item: SelectOption, selected?: boolean) => (
+        <div className="flex justify-between">
+          {item.label}
+          <div className="flex">
+            <Checkbox checked={selected} type="checkbox" />
```
Leiksa, line 37: If this `Checkbox` is just decorative (we click on the whole item to check it or uncheck it) we should pass a `tabIndex={-1}` so we don't trap the focus

lib/src/components/Select/docs/examples/icon.tsx
```diff
@@ -0,0 +1,32 @@
...
+const Example = () => {
+  const [value, setValue] = React.useState<SelectProps['value']>()
+
+  const handleChange = (newValue: SelectProps['value']) => {
+    setValue(newValue)
+  }
+
+  return (
+    <Select
+      icon={<WttjIcon alt="Welcome" color="neutral-90" />}
```
Leiksa, line 23: Same here for the `alt`

Leiksa:
Well done 🔥 Just some adjustment needs to be done 😇

lib/src/components/Select/select.module.scss
```diff
@@ -0,0 +1,336 @@
...
+    outline: 2px solid var(--select-outline-color, transparent);
+    cursor: default;
+    width: 100%;
+    transition-property: all;
+    transition-timing-function: var(--timing-primary);
+    transition-duration: var(--duration-medium);
+    appearance: none;
+    @include text-ellipsis;
+
+    &.variant-danger {
```
Leiksa: Why didn’t you nest every variant/size? 🤔 Also, the `variant` and `size` classes shouldn’t be nested inside the `.root` class. You can check how it’s written in `button.module.scss`.

```suggestion
    .variant {
         &-danger {}
         &-success {}
         ...}
```

lib/src/components/Select/index.tsx
Isaacmaamouche: renaming to initialInputValue & initialSelectedValues

lib/src/components/Select/docs/examples/allow-unselect-from-list.tsx
Isaacmaamouche: alt is to replace the use of title -> it gets turned into a aria-label and by default, icons now have a role of "img" :see-no-evil: 
I don't know what i should do here :(

Leiksa: Sorry I didn't see the PR regarding the Icon 🐙 
1. Icon should never have a role img by default as they are 100% decorative
2. As they don't have a role img -> no alt
3. Icon should be aria-hidden by default as they are purely decorative
4. If we really need for some case to make them context useful (icon only to be displayed without text next to it, we need to give more informations to the screen reader user) then we should handle it in the implementation

I will fix it in another PR 😃 cc @theo-mesnil

theo-mesnil: Thanks @Leiksa ❤️

lib/src/components/Select/select.module.scss
Isaacmaamouche: oupsie

PR #2862, opened 2025-09-19
chore: migrate AspectRatio

anucreative:
## DESCRIPTION

No breaking changes

image

## COMPATIBILITY

## QA

### DIFF
--- a/lib/src/components/AspectRatio/index.tsx
+++ b/lib/src/components/AspectRatio/index.tsx
@@ -1,20 +1,18 @@
-import type { PolymorphicProps } from '@/theme/types'
 import { classNames } from '@/utils'
 import { forwardRefWithAs } from '@/utils/forwardRefWithAs'
 
 import styles from './aspect-ratio.module.scss'
+import type { AspectRatioProps } from './types'
 
 const cx = classNames(styles)
 
-export type AspectRatioProps<T extends React.ElementType = 'div'> = PolymorphicProps<T> & {
-  ratio?: number
-}
-
 export const AspectRatio = forwardRefWithAs<AspectRatioProps, 'div'>(
-  ({ className, ratio = 4 / 3, ...rest }, ref) => {
+  ({ as, className, ratio = 4 / 3, ...rest }, ref) => {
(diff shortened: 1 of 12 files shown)


lib/src/components/AspectRatio/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { classNames } from '@/utils'
+import { forwardRefWithAs } from '@/utils/forwardRefWithAs'
+
+import styles from './aspect-ratio.module.scss'
+
+const cx = classNames(styles)
+
+export type AspectRatioProps<T extends React.ElementType = 'div'> = PolymorphicProps<T> & {
+  ratio?: number
+}
```
Isaacmaamouche: Can you add a types.ts file with an interface for `ratio` so it's picked up by the théo's doc gen thing please?

lib/src/components/AspectRatio/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const AspectRatio = forwardRefWithAs<AspectRatioProps, 'div'>(
+  ({ className, ratio = 4 / 3, ...rest }, ref) => {
+    const style = { '--aspect-ratio': ratio } as React.CSSProperties
```
Isaacmaamouche, line 11: just a question on using the `const` over  `style={{ '--aspect-ratio': ratio }}` 
- is it to make it clear that vars are being passed through ?

anucreative: We have to do it like this because the value is coming from a prop and it has infinite values. Either:

1. inline style `style = {{ aspect-ratio: ratio }}`
2. set css variables in inline styles and use in .scss

It seems we are using pattern 2

cnairi: I think what Isaac was asking is why creating a const for style, instead of just inlining it
`style={{ '--aspect-ratio': ratio } as React.CSSProperties}`

I personally have zero preferences on this topic, inlining it for one property seems ok, creating a const if it looks more readable for your is ok as well

PR #2872, opened 2025-09-22
feat: migrate Alert

theo-mesnil:


lib/src/components/Alert/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+          {!!actions && <div className="flex items-center gap-sm">{actions}</div>}
+        </div>
+      </div>
+    )
+  }
+)
+
+// We need this component to check its existence in <Alert> and to allow users to add Button in <Alert> content
+const AlertButton = forwardRef<HTMLButtonElement, Omit<ButtonProps, 'size'>>(
+  ({ variant = 'secondary', ...props }, ref) => (
```
anucreative, line 95: It would be better / cleaner to override the variant here than in `cloneActions`

lib/src/components/Alert/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+            size="xs"
+            top="sm"
+          />
+        )}
+        {showIcon ? (
+          <VariantIcon className="self-start" size={size} variant={defaultVariantIcon} />
+        ) : null}
+        <div className={cx('content')}>
+          <Text className="flex-1">{content}</Text>
+          {!!actions && <div className="flex items-center gap-sm">{actions}</div>}
```
anucreative: We should try not to use Tailwind _in_ our components. This could easily be a div with its own class.

lib/src/components/Alert/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+              cloneActions(nestedChild)
+            ),
+          })
+        }
+      }
+
+      return child
+    }
+
+    const actions = React.isValidElement(cta) ? cloneActions(cta) : cta
```
anucreative, line 61: Do we really need all this ? Or can we just make use of `Alert.AlertButton` etc. For the `size` it would be simple in CSS to set / override the size of child buttons etc without cloning elements.

anucreative (commented):
On a besoin des styles dans `global.css`. 

Et je question le besoin de tous ces `cloneElement` etc

lib/src/components/Alert/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+            right="sm"
+            size="xs"
+            top="sm"
+          />
+        )}
+        {showIcon ? (
+          <VariantIcon className="self-start" size={size} variant={defaultVariantIcon} />
+        ) : null}
+        <div className={cx('content')}>
+          <Text className="flex-1">{content}</Text>
```
anucreative: If we pass content other than text we get an error `div cannot be a descendant of p`

lib/src/components/Alert/index.tsx
theo-mesnil: It's better to pass the size to the button component to avoid rewriting in CSS. I think this is OK for now, and we can update it when we've finished migrating the last components.

lib/src/components/Alert/index.tsx
theo-mesnil: The child can be only a text, or can have a link but not another paragraph, it's specifications from design :)

lib/src/components/Alert/index.tsx
theo-mesnil: How would you do it without passing the component in your alert in your Alert.Button? We need to force ai variant

lib/src/components/Alert/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+const cx = classNames(alertStyles)
+
+const AlertComponent = forwardRef<HTMLDivElement, AlertProps>(
+  (
+    {
+      children,
+      className,
+      cta,
+      dataTestId,
```
cnairi: What should we do about the dataTestId props? I thought we were trying to use data-testid instead?

theo-mesnil: Yes 👍🏻

lib/src/components/Alert/index.tsx
anucreative: We use `Alert` component for the Growl / Snackbar components. @diez-wttj : Might need to update your PR to make sure we're only sending text.

lib/src/components/Alert/index.tsx
anucreative: Would this work ? 

```
<Alert
  cta={
    <>
      <Alert.Button variant="secondary-ai">Button</Alert.Button>
      <Alert.SecondaryButton>Button</Alert.SecondaryButton>
    </>
  }>
  <span>My content</span>
</Alert>
```

theo-mesnil: We don't want this because we want for Alert.Button on ia force variant to avoid mistake, that why we keep for now cloneElements

lib/src/components/Alert/index.tsx
theo-mesnil: I'll change to a div for Text for now, but also we need to rework this :D

PR #2875, opened 2025-09-22
feat: migrate close button

StephaneRob:


lib/src/components/CloseButton/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import closeButtonStyles from './close-button.module.scss'
+
+const cx = classNames(closeButtonStyles)
+
+export type CloseButtonProps = ButtonProps
+
+export const CloseButton = forwardRef<HTMLButtonElement, CloseButtonProps>((props, ref) => {
+  return (
+    <Button
+      className={cx('root')}
```
RobelTekle:
```suggestion
export const CloseButton = forwardRef<HTMLButtonElement, CloseButtonProps>(({ className, ...props}, ref) => {
  return (
    <Button
      className={cx('root', className)}
```

lib/src/components/CloseButton/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+      ref={ref}
+      shape="circle"
+      size="sm"
+      title="Close"
```
Leiksa:
```suggestion
      aria-label="Close"
```
The prop `title` doesn't give any context for a11y, we should use `aria-label`

lib/src/components/CloseButton/index.tsx
```diff
@@ -0,0 +1,29 @@
+import { forwardRef } from 'react'
+
+import { classNames } from '@/utils'
+
+import { Button } from '../Button'
```
Leiksa:
```suggestion
import { Button } from '@/components/Button'
```

lib/src/components/CloseButton/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import type { ButtonProps } from '../Button/types'
```
Leiksa:
```suggestion
import type { ButtonProps } from '@/components/Button/types'
```

lib/src/components/CloseButton/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { CrossIcon } from '../Icon'
```
Leiksa:
```suggestion
import { CrossIcon } from '@/components/Icon'
```

Leiksa (approved):
Some minor adjustments needed 😄

PR #2876, opened 2025-09-22
chore: migrate Breadcrumb component

SimonDegraeve:
## DESCRIPTION

Migrate the `Breadcrumb` component. There are no props change. I also extracted the mobile logic gradient in a hook.

image

## COMPATIBILITY

## QA


lib/src/components/Breadcrumb/index.test.tsx
```diff
@@ -0,0 +1,63 @@
+import { render } from '../../../tests'
```
theo-mesnil:
```suggestion
import { render } from '@tests'
```

lib/src/components/Breadcrumb/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+          {clones.reverse()}
+        </ol>
+
+        {endGradient}
+      </nav>
+    )
+  }
+)
+
+BreadcrumbComponent.displayName = 'Breadcrumb'
```
theo-mesnil: Pas besoin à supprimer

lib/src/components/Breadcrumb/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react'
+
+import { primitives } from '@/theme/tokens'
+import { classNames } from '@/utils'
+import { RightIcon } from '@old/Icons'
+import { clamp, hexToRGBA, throttle } from '@old/utils'
```
theo-mesnil: Il faut voir si on veut ajouter aux utils si on a en besoin,

lib/src/components/Breadcrumb/Item.tsx
```diff
@@ -0,0 +1,36 @@
+import { forwardRef } from 'react'
```
theo-mesnil, line 1: rename to Item.tsx

lib/src/components/Breadcrumb/types.ts
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable perfectionist/sort-modules */
+import type { ComponentPropsWithRef, HTMLAttributes } from 'react'
+
+import type { ThemeColorTokens } from '@old/theme'
```
theo-mesnil: Il ne faut plus utiliser cela, plus se dire que c'est du css ?

lib/src/components/Breadcrumb/breadcrumb.module.scss
```diff
@@ -0,0 +1,43 @@
...
+    right: 0;
+    transform-origin: right;
+  }
+
+  .startGradient,
+  .endGradient {
+    position: absolute;
+    bottom: 0;
+    top: 0;
+    width: 30;
```
cnairi: 30 is not a valid value

lib/src/components/Breadcrumb/breadcrumb.module.scss
```diff
@@ -0,0 +1,43 @@
...
+    letter-spacing: var(--letter-spacing-subtitle-sm);
+    line-height: var(--line-height-subtitle-sm);
+    text-transform: uppercase;
+  }
+
+  .list {
+    display: inline-flex;
+    align-items: center;
+    max-width: 100%;
+    height: 100%;
```
cnairi: height: 100% is already defined on root

lib/src/components/Breadcrumb/docs/examples/child-disabled.tsx
```diff
@@ -0,0 +1,13 @@
+import { Breadcrumb } from '@old/Breadcrumb'
```
cnairi: You shouldn't import the old Breadcrumb component in your doc

lib/src/components/Breadcrumb/docs/examples/separator.tsx
```diff
@@ -0,0 +1,22 @@
...
+        <Breadcrumb.Item href="/">Components</Breadcrumb.Item>
+        <Breadcrumb.Item>Breadcrumb</Breadcrumb.Item>
+      </Breadcrumb>
+
+      <Breadcrumb separator="-">
+        <Breadcrumb.Item href="/">Introduction</Breadcrumb.Item>
+        <Breadcrumb.Item href="/">Components</Breadcrumb.Item>
+        <Breadcrumb.Item>Breadcrumb</Breadcrumb.Item>
+      </Breadcrumb>
+    </Stack>
```
cnairi: Stack should be replaced with <div className="flex"> as it's going to be deprecated

lib/src/components/Breadcrumb/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react'
+
+import { primitives } from '@/theme/tokens'
+import { classNames } from '@/utils'
+import { RightIcon } from '@old/Icons'
```
cnairi: Replace old RightIcon with the new one please `import { RightIcon } from '@/components/Icon'`

lib/src/components/Breadcrumb/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { Item } from './item'
+import type { BreadcrumbProps } from './types'
+
+const cx = classNames(breadcrumbStyles)
+
+export const BreadcrumbComponent = forwardRef<HTMLDivElement, BreadcrumbProps>(
+  (
+    {
+      children,
+      gradientBackground = 'neutral-10',
```
cnairi: Could we remove the gradientBackground? It's only used in welcome-home

lib/src/components/Breadcrumb/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+      return cloneElement(child, {
+        isActive,
+        key: `breadcrumb-${index}`,
+        separator: isLastChild ? undefined : separator,
+        ...child.props,
+      })
+    })
+
+    // Add scroll gradient on mobile
+    const { endGradient, listRef, onListScroll, startGradient } = useGradient(gradientBackground)
```
cnairi: I would completely get ridd off `gradientBackground`, @theo-mesnil what do you think? (only used in wh)

lib/src/components/Breadcrumb/item.tsx
```diff
@@ -0,0 +1,36 @@
...
+import breadcrumbItemStyles from './item.module.scss'
+import type { BreadcrumbItemProps } from './types'
+
+const cx = classNames(breadcrumbItemStyles)
+
+/**
+ * @name Breadcrumb.Item
+ */
+export const Item = forwardRef<HTMLAnchorElement, BreadcrumbItemProps>(
+  ({ children, dataTestId, isActive, separator, ...rest }, ref) => {
```
cnairi: I think we are trying to use 'data-testid' instead of dataTestId

lib/src/components/Breadcrumb/docs/examples/separator.tsx
theo-mesnil: Use <div className /> instead

lib/src/components/Breadcrumb/index.tsx
theo-mesnil: Hum are you sure? We need to keep this on mobile i think

SimonDegraeve: So, should I remove the gradient ?

SimonDegraeve:
Thank you for the feedback, I will do the changes this morning

lib/src/components/Breadcrumb/index.tsx
theo-mesnil: For me now we need to keep it for mobile style, this can be rework on belonging follows the new design maybe

SimonDegraeve: Maybe I can keep the gradient but remove the customizable color, what do you think ?

cnairi: Yes, sorry I didn't realize you had a default value on useGradient, even without gradientBackground!!

lib/src/components/Breadcrumb/breadcrumb.module.scss
```diff
@@ -0,0 +1,42 @@
...
+    transform-origin: right;
+  }
+
+  .startGradient,
+  .endGradient {
+    position: absolute;
+    bottom: 0;
+    top: 0;
+    width: var(--spacing-xxl);
+  }
```
theo-mesnil: Do we want to use camelCase or snake-case ?
cc @cnairi

PR #2886, opened 2025-09-23
chore: migrate FileUpload & Files

baptistetruchet:
I removed one example that contains the `Field` component


lib/src/components/FileUpload/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+      accept = DEFAULT_FILE_TYPES,
+      children,
+      disabled,
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      draggable,
+      handleAddFile,
+      handleRemoveFile,
+      // Previously pass to input props as `maxSize` but not a valid attribute
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      maxSize = DEFAULT_MAX_FILE_SIZE,
```
baptistetruchet: Should I remove it completely? If I am not mistaken, it had no effect on the input

lib/src/components/FileUpload/index.tsx
```diff
@@ -0,0 +1,158 @@
+import { forwardRef, useEffect, useImperativeHandle, useRef, useState } from 'react'
+
+import { Button } from '@/components/Button'
+import { getFileIcon, getFileName, getFileSize } from '@/components/Files'
+import { Tag } from '@/components/Tag'
+import { classNames } from '@/utils'
+import { createEvent } from '@/utils/create-event'
```
cnairi, line 7: You cannont `import { classNames, createEvent } from '@/utils'`?

lib/src/components/FileUpload/index.tsx
cnairi: - FileUpload + maxSize is only used twice (in wttj-front)
- maxSize was not working anyway
- You can either handle it (as is the case in FileDrop for instance), or just remove it as it's not used anyway

Same thing for the draggable props, why do we have it if it's not used? I would remove it completely

lib/src/components/Files/types.ts
```diff
@@ -0,0 +1,140 @@
+// https://stackoverflow.com/questions/66500729/union-type-of-keyof-and-string-with-working-autocompletion
+export type Type = (string & {}) | Types[keyof Types]
```
cnairi: I think this utils could answer this need `lib/src/old/utils/ts-utils.ts`

cnairi (commented):
i think you can also keep the example with the old field, it will be replaced when field is done?

lib/src/components/Files/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  VideoIcon,
+  XlsIcon,
+  XlsxIcon,
+  ZipIcon,
+} from '@/components/Icon'
+import { formatBytes } from '@/utils/format-bytes'
+
+import { types } from './types'
+
+export type FileType = File | string
```
Isaacmaamouche: all types should be in types.ts

Isaacmaamouche (approved):

lib/src/components/FileUpload/index.tsx
baptistetruchet: I will remove it 👍

lib/src/components/FileUpload/index.tsx
baptistetruchet: It's currently not exported from utils index. I thought we tried to avoid barrel files

PR #2887, opened 2025-09-23
Chore: migrate modal and asset modal to tailwind [WUI-150]

Isaacmaamouche:
It's a big component with a lot of small files

#### FYI 
[X] It was dependent on the swiper to style swipers inside asset modals
     -> I used a fairly ok selector but i don't know how much it covers all cases **-> Selector is okay for now** 
     -> I left a TODO
     
[X] I didn't forward className anywhere ever 
     -> it uses the composition pattern + it has render fonction for overriding **-> OKAY**

[X] I left the code i wanted to dead lib/src/components/Modal/components/Content/index.tsx
     -> Instead I migrated it to use css property names and removed styled props names **-> OKAY for now**

#### Open questions

[ ] Since we can't use vrs for breakpoint, I left a comment next to them for in case of futur changes
     -> Any recommendations ?
     -> **try SASS variables and SASS functions**

[X] Folder architecture is all over the place, I couldn't find something I like 
     -> Modal/index.tsx has a bunch of small components 
     -> Assets/XXX has a bunch a small files 
     -> **REWORKED  to -> Modal/components/--Entity--.tsx**

[X] I am not sure about any of the typing I did when it comes to type vs interface **-> OKAY**

[X] text-center no longer works because icons are display: block
    -> should be inline-block ?
    -> in the meantime, i made the header a display: flex -> maybe good enough/ better ? **-> OKAY**

[X] I used a css file per entity: Header, Footer, etc 
     -> But only one for the AssetModal 
     -> Which one is prefered ? 
     **-> ONLY ONE**

[X] For absolute sized element, like the modal, we use rem over px.
     -> For fonts stuff I get, it scales with the user settings, but for UI element, should we ?
     -> What do you think about using a toRem function like we use to have ? [ Article here](https://dev.to/nikolab/convert-px-to-rem-using-sass-3-methods-4ep2)
      -> **WE GO WITH REM**


lib/src/components/Modal/Assets/AssetBackdrop.tsx
```diff
@@ -0,0 +1,32 @@
+import { DialogDismiss } from '@ariakit/react'
+import { forwardRef } from 'react'
+
+import { CrossIcon } from '@/components/Icon'
+import type { BackdropProps } from '@/components/Modal/types'
```
anucreative: Should this be a relative path ?

lib/src/components/Modal/components/Header.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const cx = classNames(modalStyles)
+
+/**
+ * @name Modal.Header
+ */
+export const Header = forwardRef<HTMLDivElement, HeaderProps>(
+  ({ icon, subtitle, title, ...rest }, ref) => {
+    //FIXME text-center no longer works because icons are display block
+    // shold be nline block or should i make the header a display flex
```
anucreative: I would use `display: flex`

lib/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+  ) => {
+    return (
+      <Dialog
+        aria-label={ariaLabel}
+        backdrop={<Backdrop backdrop={backdrop} hideOnInteractOutside={hideOnInteractOutside} />}
+        hideOnInteractOutside={hideOnInteractOutside}
+        ref={ref}
+        render={props =>
+          As ? <As {...props} /> : <div className={cx('root', `size-${size}`)} {...props} />
+        }
```
anucreative: I've been using the pattern:

```
const Element = as || 'div'

…

render={props => <Element className={cx('root', `size-${size}`)} {...props} />}
```

Would this work in your case ?

lib/src/components/Modal/components/Content.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { Children, cloneElement, forwardRef, useEffect, useMemo, useState } from 'react'
+
+import { Close } from '@/components/Modal/components/Close'
```
anucreative: Should this be relative ?

lib/src/components/Modal/index.tsx
Isaacmaamouche: The classNames would be passed to As, it could disrupt the user's passed Element :/

lib/src/components/Modal/Assets/AssetBackdrop.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { classNames } from '@/utils'
+
+import modalStyles from './assets.module.scss'
+import { CloseButton } from './CloseButton'
+const cx = classNames(modalStyles)
```
cnairi, line 12: no big deal but could you add a line break between your imports and the const cx declaration please?

lib/src/components/Modal/Assets/Iframe.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import modalStyles from './assets.module.scss'
+import type { IframeProps } from './types'
+
+const cx = classNames(modalStyles)
+
+export const Iframe = ({ children }: IframeProps) => {
+  return (
+    <div className={cx('root', 'iframe-wrapper')}>
+      <div className="h-full mx-0 my-auto aspect-[calc(16/9)]">{children}</div>
```
cnairi: aspect-[calc(16/9) = aspect-video (https://tailwindcss.com/docs/aspect-ratio#using-a-video-aspect-ratio)

lib/src/components/Modal/Assets/types.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export type AssetWithTitleProps = MergeProps<
+  AssetWithTitleOptions,
+  React.HTMLAttributes<HTMLDivElement>
+>
+
+export type BackdropProps = Pick<DialogOptions, 'hideOnInteractOutside'>
+
+export type IframeProps = {
+  children: JSX.Element
```
cnairi: JSX.Element or React.ReactNode?

lib/src/components/Modal/components/Body.tsx
```diff
@@ -0,0 +1,13 @@
...
+import type { BodyProps } from '@/components/Modal/types'
+import { classNames } from '@/utils'
+
+const cx = classNames(modalStyles)
+
+export const Body = forwardRef<HTMLElement, BodyProps>((props, ref) => {
+  return <section className={cx('body')} ref={ref} {...props} />
+})
+
+Body.displayName = 'Body'
```
cnairi, line 13: You can remove the displayName

lib/src/components/Modal/components/Footer.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * @name Modal.Footer
+ */
+export const Footer = forwardRef<HTMLDivElement, FooterProps>(
+  ({ children, information, ...rest }, ref) => {
+    return (
+      <footer className={cx('footer')} ref={ref} {...rest}>
+        {children ? <div className={cx('footer-children-wrapper')}>{children}</div> : null}
+        {information ? (
+          <div className={cx('footer-information')}>
+            <Text className="font-bold text-(--color-neutral-90)" variant="subtitle-sm">
```
cnairi, line 21: text-neutral-90

lib/src/components/Modal/components/Footer.tsx
```diff
@@ -0,0 +1,33 @@
...
+              {information.title}
+            </Text>
+            <Text className="text-(--color-neutral-90) mb-0 mt-(--spacing-md)" variant="sm">
```
cnairi, line 24: text-neutral-90
mt-md

lib/src/components/Modal/components/Header.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * @name Modal.Header
+ */
+export const Header = forwardRef<HTMLDivElement, HeaderProps>(
+  ({ icon, subtitle, title, ...rest }, ref) => {
+    return (
+      <header className={cx('header')} ref={ref} {...rest}>
+        <Close isOnHeader />
+        {icon}
+        <Text
+          className={`${subtitle && 'mb-(--spacing-lg)'} ${icon && 'mt-(--spacing-xl)'}`}
```
cnairi, line 23: mb-lg
mt-xl

lib/src/components/Modal/components/Header.tsx
```diff
@@ -0,0 +1,32 @@
...
+        >
+          {title}
+        </Text>
+        {subtitle ? <Text className={cx('header-subtitle')}>{subtitle}</Text> : null}
+      </header>
+    )
+  }
+)
+
+Header.displayName = 'Header'
```
cnairi, line 34: You can remove the display name as well

lib/src/components/Modal/docs/examples/header-footer.tsx
```diff
@@ -0,0 +1,80 @@
+import { Button } from '@/components/Button'
+import { ActionsIcon } from '@/components/Icon'
+import { Modal, useModal } from '@/components/Modal'
+import { Box } from '@old/Box'
```
cnairi: You can remove the Box now

lib/src/components/Modal/docs/examples/header-footer.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </Modal.Trigger>
+      <Modal ariaLabel="example" store={modal}>
+        <Modal.Content store={modal}>
+          <Modal.Header icon={icon} subtitle={subtitle} title={title} />
+          <Modal.Body>
+            Praesent sit amet quam ac velit faucibus dapibus. Quisque sapien ligula, rutrum quis
+            aliquam nec, convallis sit amet erat. Mauris auctor blandit porta.
+          </Modal.Body>
+          <Modal.Footer information={informations}>
+            <Box w="100%">
```
cnairi:
```suggestion
            <div className="w-full">
```

lib/src/components/Modal/docs/examples/header-footer.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </Modal.Trigger>
+      <Modal ariaLabel="example-2" store={modal2}>
+        <Modal.Content store={modal2}>
+          <Modal.Header subtitle={subtitle} title={title} />
+          <Modal.Body>
+            Praesent sit amet quam ac velit faucibus dapibus. Quisque sapien ligula, rutrum quis
+            aliquam nec, convallis sit amet erat. Mauris auctor blandit porta.
+          </Modal.Body>
+          <Modal.Footer>
+            <Box w="100%">
```
cnairi:
```suggestion
            <div className="w-full">
```

lib/src/components/Modal/modal.module.scss
```diff
@@ -0,0 +1,136 @@
...
+    &.hideOnInteractOutside {
+      cursor: pointer;
+    }
+
+    &[data-enter] {
+      opacity: 1;
+    }
+  }
+
+  .root {
```
cnairi, line 4: I would put the root above the backdrop (no big deal)

lib/src/components/Modal/Assets/AssetWithTitle.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Text } from '@/components/Text'
+import { classNames } from '@/utils'
+
+import modalStyles from './assets.module.scss'
+import { AssetTitle } from './AssetTitle'
+import type { AssetWithTitleProps } from './types'
+
+const cx = classNames(modalStyles)
+
+export const AssetWithTitle = forwardRef<HTMLDivElement, AssetWithTitleProps>(
```
Leiksa, line 12:
```suggestion
export const AssetWithTitle = forwardRefWithAs<AssetWithTitleProps,'div'>(
```

lib/src/components/Modal/Assets/CloseButton.tsx
```diff
@@ -0,0 +1,14 @@
+import { Button } from '@/components/Button'
+import type { ButtonProps } from '@/components/Button/types'
+import { classNames } from '@/utils'
+
+import modalStyles from './assets.module.scss'
+const cx = classNames(modalStyles)
+
+export const CloseButton = ({ children, ...rest }: ButtonProps) => {
```
Leiksa, line 8:
```suggestion
export const CloseButton = forwardRefWithAs<ButtonProps, 'button'>((props, ref) => {
```

lib/src/components/Modal/Assets/Dialog.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { classNames } from '@/utils'
+
+import modalStyles from './assets.module.scss'
+const cx = classNames(modalStyles)
+
+export const Dialog = ({
+  children,
+  ...rest
+}: React.PropsWithChildren<HTMLAttributes<HTMLDivElement>>) => {
```
Leiksa, line 12: same here (forwardRefWithAs)

lib/src/components/Modal/Assets/Iframe.tsx
Leiksa: IMO we should not use tailwind here but our scss as we can easily handle it -> `aspect-ratio: 16/9`

lib/src/components/Modal/Assets/Iframe.tsx
```diff
@@ -0,0 +1,14 @@
+import { classNames } from '@/utils'
+
+import modalStyles from './assets.module.scss'
+import type { IframeProps } from './types'
+
+const cx = classNames(modalStyles)
+
+export const Iframe = ({ children }: IframeProps) => {
```
Leiksa: forwardRefWithAs 🫢

lib/src/components/Modal/Assets/types.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface AssetWithTitleOptions {
+  children: React.ReactNode
+  customContent?: JSX.Element | string
+  subtitle?: JSX.Element | string
+  title?: JSX.Element | string
+}
+
+export type AssetWithTitleProps = MergeProps<
+  AssetWithTitleOptions,
+  React.HTMLAttributes<HTMLDivElement>
```
Leiksa:
```suggestion
  ComponentProps<'div'>
```

lib/src/components/Modal/components/Body.tsx
```diff
@@ -0,0 +1,13 @@
+import { forwardRef } from 'react'
+
+import modalStyles from '@/components/Modal/modal.module.scss'
+import type { BodyProps } from '@/components/Modal/types'
+import { classNames } from '@/utils'
+
+const cx = classNames(modalStyles)
+
+export const Body = forwardRef<HTMLElement, BodyProps>((props, ref) => {
```
Leiksa, line 9: forwardRefWithAs 🫢

lib/src/components/Modal/components/Body.tsx
Leiksa: We should keep it. Sometimes having HoC on top of component rendering can lead to React not infering the correct component name so we should in general put it everywhere we are using `memo`, `forwardRef`, `forwardRefWithAs`

lib/src/components/Modal/components/Close.tsx
```diff
@@ -0,0 +1,17 @@
+import { DialogDismiss } from '@ariakit/react'
+
+import { CloseButton } from '@/components/CloseButton'
+import type { CloseProps } from '@/components/Modal/types'
+
+export const Close = ({ isOnHeader, ...rest }: CloseProps) => {
```
Leiksa, line 7: ForwardRefWithAs 🫢 (Not really usefull here but we should at least be sure that the `rest` contains every props that can handle the button)

lib/src/components/Modal/components/Content.tsx
```diff
@@ -0,0 +1,85 @@
...
+import type { ContentProps } from '@/components/Modal/types'
+
+/**
+ * @name Modal.Content
+ */
+export const Content = forwardRef<HTMLDivElement, ContentProps>(
```
Leiksa, line 10: forwardRefWithAs 🫢 (sorry to be redundant 😇 )

lib/src/components/Modal/components/Content.tsx
```diff
@@ -0,0 +1,85 @@
...
+            const componentType = child.type as React.ComponentClass | React.FunctionComponent
+            return componentType.displayName ?? componentType.name ?? ''
+          }
+          return ''
+        }),
+      [children]
+    )
+    const shouldShowCloseButton = !components.includes('Header') && withClosingButton
+
+    const getStyles = (name?: string) => {
```
Leiksa, line 29: maybe we can handle all of that thanks CSS now since we can use the `:has` selector

lib/src/components/Modal/components/Footer.tsx
```diff
@@ -0,0 +1,33 @@
...
+import type { FooterProps } from '@/components/Modal/types'
+import { Text } from '@/components/Text'
+import { classNames } from '@/utils/classNames'
+
+const cx = classNames(modalStyles)
+
+/**
+ * @name Modal.Footer
+ */
+export const Footer = forwardRef<HTMLDivElement, FooterProps>(
```
Leiksa, line 14: forwardRefWithAs 👀

lib/src/components/Modal/components/Footer.tsx
Leiksa: why do we use tailwind in our components? Shouldn't we just use SCSS/CSS and avoid putting tailwind directly inside of them to not be that much dependent?

lib/src/components/Modal/components/Header.tsx
```diff
@@ -0,0 +1,32 @@
...
+import type { HeaderProps } from '@/components/Modal/types'
+import { Text } from '@/components/Text'
+import { classNames } from '@/utils/classNames'
+
+const cx = classNames(modalStyles)
+
+/**
+ * @name Modal.Header
+ */
+export const Header = forwardRef<HTMLDivElement, HeaderProps>(
```
Leiksa, line 16: You know what I mean 🙈

lib/src/components/Modal/components/Trigger.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import type { TriggerProps } from '@/components/Modal/types'
+import { forwardRefWithAs } from '@/utils'
+
+export const Trigger = forwardRefWithAs<TriggerProps, 'button'>(
+  ({ as: As, store, ...rest }, ref) => {
+    return (
+      <DialogDisclosure
+        ref={ref}
+        render={As ? props => <As {...props} /> : undefined}
```
Leiksa: No need to pass the props as it will handle it by default. It is only useful in some case you want to override some of them
```suggestion
        render={As ? <As /> : undefined}
```

lib/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    ...storeOptions,
+  })
+
+  return dialog
+}
+
+const ModalComponent = forwardRefWithAs<ModalProps, 'div'>(
+  (
+    {
+      ariaLabel,
```
Leiksa: no need to extract it as it will be forwarded in the `rest` spread

lib/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    ref
+  ) => {
+    return (
+      <Dialog
+        aria-label={ariaLabel}
+        backdrop={<Backdrop backdrop={backdrop} hideOnInteractOutside={hideOnInteractOutside} />}
+        hideOnInteractOutside={hideOnInteractOutside}
+        ref={ref}
+        render={props =>
+          As ? <As {...props} /> : <div className={cx('root', `size-${size}`)} {...props} />
```
Leiksa: Same here for the props spreading. Also you still need to pass the `className` in order to make it overridable (don't forget to apply it to the `Dialog` element as it must be forwarded to the `As`
```suggestion
          As ? <As /> : <div className={cx('root', `size-${size}`, className)}/>
```

lib/src/components/Modal/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  ariaLabel: string
+  children: React.ReactElement
+  size?: Size
+}
+
+type BaseDialogOptions = Omit<DialogOptions<'div'>, 'as'>
+
+export type ModalProps = MergeProps<
+  BaseDialogOptions & ModalOptions,
+  HTMLAttributes<HTMLDivElement>
```
Leiksa:
```suggestion
  ComponentProps<'div'>
```

lib/src/components/Modal/Assets/AssetWithTitle.tsx
Isaacmaamouche: Why the `forwardRefWithAs` ?

lib/src/components/Modal/Assets/CloseButton.tsx
Isaacmaamouche: We should not allow this button to be messed with imho 🙈

lib/src/components/Modal/Assets/Iframe.tsx
Isaacmaamouche: Diane's right + i don't like that with a two div for no reasons. I'll try to merge the two in `.iframe-wrapper`

lib/src/components/Modal/Assets/types.ts
Isaacmaamouche: Image

Isaacmaamouche: I will delete this type and use `React.PropsWithChildren` on the component itself

lib/src/components/Modal/Assets/types.ts
Isaacmaamouche: @cnairi I think we another great discussion on our hands: 
- ComponentProps is the right call, it includes all of HTMLAttributes and React specific props 
- We're using all them in many ways all over the codebase :s https://github.com/search?q=repo%3AWTTJ%2Fwelcome-ui%20HTMLAttributes&type=code

According to the voices in my head: 

In React, both `HTMLAttributes<HTMLDivElement>` and `ComponentProps<'div'>` can be used to type the props of a component to match the props a `<div>` element would take, but they have slightly different use cases:

### **1. `HTMLAttributes<HTMLDivElement>`**
- This type includes all the standard HTML attributes that a `<div>` element can accept, such as `className`, `id`, `style`, etc.
- It is more specific to the `<div>` element and does not include React-specific props like `ref`.

### **2. `ComponentProps<'div'>`**
- This type is broader and includes all the props that a `<div>` element can accept, including React-specific props like `ref`.
- It is equivalent to `HTMLAttributes<HTMLDivElement>` combined with `ClassAttributes<HTMLDivElement>`.

### **Which is better?**
- If you need to include `ref` handling in your component, **`ComponentProps<'div'>`** is the better choice because it includes the `ref` prop.
- If you don't need `ref` handling, **`HTMLAttributes<HTMLDivElement>`** is sufficient.

### **In your case:**
It's usually better to use **`ComponentPropsWithRef`** or **`ComponentPropsWithoutRef`** instead of this type, as they let you be explicit about whether or not to include the ref prop.

lib/src/components/Modal/components/Body.tsx
Isaacmaamouche: displayNames are used in Content.tsx for dark shaman magic styling stuff

lib/src/components/Modal/components/Content.tsx
Isaacmaamouche: we decided to do it later

lib/src/components/Modal/components/Footer.tsx
Isaacmaamouche: `text-neutral-90` because tailwind adopts our theme values ? Or by luck w-ui and tailwind uses the same values ?

lib/src/components/Modal/index.tsx
Isaacmaamouche: > don't forget to apply it to the Dialog

If we're passing our own component `As` we give it our own classNames. Wouldn't passing `Dialog`'s classNames to`As` duplicate them ?

Leiksa: In theory, AriaKit merged the props from the Dialog and the one from the render prop element

lib/src/components/Modal/components/Footer.tsx
cnairi: tailwind and wui uses the same values!

PR #2890, opened 2025-09-23
[WUI-224] fix(icon): a11y, remove alt & add aria-hidden by default

Leiksa:
## DESCRIPTION

This pull request refactors the `Icon` component and its usage throughout the codebase to improve accessibility and standardize its API. The main changes include replacing the `alt` prop with `aria-label`, updating the component signature and types, and adjusting tests and usages accordingly. Additionally, the tests for both `Icon` and `Alert` components have been updated to reflect these changes and improve coverage.

**Icon component refactor and accessibility improvements:**

* Updated `Icon.tsx` to use `aria-label` instead of `alt`, set `aria-hidden` when no label is provided, and conditionally set `role="img"` only when `aria-label` is present. The `<title>` element was also removed. (`[lib/src/components/Icon/Icon.tsxL11-L25](diffhunk://#diff-7a06eb7185ddd71d19b5ee3f6abeee9830978c72341ffe66a636912101fa0d53L11-L25)`)
* Refactored `Icon.types.ts` to use a new `ForwardRefProps` type and removed the `alt` prop from `IconProps`, aligning the types with the new API. (`[[1]](diffhunk://#diff-5b7161b6a8602ba2205b85102bdded46091119755091117fd00f8c42fcbc03d7L1-R8)`, `[[2]](diffhunk://#diff-5b7161b6a8602ba2205b85102bdded46091119755091117fd00f8c42fcbc03d7L19-R19)`)


lib/src/components/Alert/index.test.tsx
```diff
@@ -14,15 +14,16 @@ describe('<Alert>', () => {
...
+    const { container } = render(
       <Alert>
         <Alert.Title data-testid="alert-title">title</Alert.Title>
         {content}
       </Alert>
     )
     const alertTitle = screen.getByTestId('alert-title')
 
-    expect(screen.getByText('Promote')).toBeInTheDocument()
+    // TODO: How do we check the "Promote" content as we don't have an aria-label anymore?
```
cnairi: Not the best idea but you could test if the icon variant has the icon-variant class, but not the warning / success / info / error variant classes

lib/src/components/Select/tests/index.test.tsx
```diff
@@ -341,21 +341,21 @@ describe('<Select>', () => {
...
+    expect(icon).toBeInTheDocument()
   })
 
   it('<Select icon> shows icon', () => {
-    const { container } = render(
+    render(
       <Select
         dataTestId="select"
-        icon={<AvatarIcon alt="Avatar" color="neutral-80" />}
+        icon={<AvatarIcon color="neutral-80" data-testid="avatar-icon" />}
```
cnairi: Is it normal that color still work?

Shouldn't it be a className?

lib/src/components/Select/docs/examples/icon.tsx
```diff
@@ -20,7 +20,7 @@ const Example = () => {
 
   return (
     <Select
-      icon={<WttjIcon alt="Welcome" color="neutral-90" />}
+      icon={<WttjIcon color="neutral-90" />}
```
cnairi: I don't think the color props should be accepted anymore

lib/src/components/Select/tests/index.test.tsx
Leiksa: You are right, I didn't pay attention to this 🙈

PR #2895, opened 2025-09-23
chore: migrate drawer

anucreative:

image

## COMPATIBILITY

## QA

### DIFF
--- a/lib/src/components/Drawer/AssetDrawer/IconBlock.tsx
+++ b/lib/src/components/Drawer/AssetDrawer/IconBlock.tsx
@@ -8,7 +8,7 @@ const cx = classNames(styles)
 export const IconBlock = ({ icon: Icon, size = 'md' }: IconBlockProps) => {
   return (
     <div className={cx('icon-block', `size-${size}`)}>
-      <Icon color="neutral-90" size={size} />
+      <Icon className="text-neutral-90" size={size} />
     </div>
   )
 }
--- a/lib/src/components/Drawer/AssetDrawer/asset-drawer.module.scss
+++ b/lib/src/components/Drawer/AssetDrawer/asset-drawer.module.scss
@@ -1,15 +1,10 @@
-@use '../../../../src/utils/breakpoints' as breakpoints;
+@use '../../../utils/breakpoints' as breakpoints;
 
 @layer components {
   .asset-drawer {
-    height: 100%;
     overflow: hidden;
     border-top-left-radius: var(--radius-2xl);
     border-top-right-radius: var(--radius-2xl);
-
-    @media screen and (min-width: breakpoints.$breakpoint-md) {
-      height: calc(100% - 3rem);
-    }
   }
 
   .content-wrapper {
@@ -85,15 +80,17 @@
     background-color: var(--color-beige-20);
     transition: background-color var(--timing-primary);
 
+    --icon-block-size: var(--spacing-3xl); /* 48px */
+
     width: var(--icon-block-size);
     height: var(--icon-block-size);
 
     &.size-md {
-      --icon-block-size: 48px;
+      --icon-block-size: var(--spacing-3xl); /* 48px */
     }
 
     &.size-sm {
-      --icon-block-size: 32px;
+      --icon-block-size: var(--spacing-xxl); /* 32px */
     }
   }
 }
--- a/lib/src/components/Drawer/AssetDrawer/index.tsx
+++ b/lib/src/components/Drawer/AssetDrawer/index.tsx
@@ -1,3 +1,5 @@
+import { useCallback } from 'react'
+
 import { classNames } from '@/utils'
 import { forwardRefWithAs } from '@/utils/forwardRefWithAs'
 
@@ -20,20 +22,26 @@ export const AssetDrawerComponent = forwardRefWithAs<AssetDrawerProps, 'div'>(
     },
     ref
   ) => {
-    const getPersistentElements = () =>
-      Array.from(
-        parentGetPersistentElements
-          ? parentGetPersistentElements()
-          : document.querySelectorAll('[data-wui-persistent]')
-      )
+    const getPersistentElements = useCallback(
+      () =>
+        Array.from(
+          parentGetPersistentElements
+            ? parentGetPersistentElements()
+            : document.querySelectorAll('[data-wui-persistent]')
+        ),
+      [parentGetPersistentElements]
+    )
 
-    const hideOnInteractOutsideFn = (event: Event) => {
-      const target = event.target as HTMLElement
-      const isTargetWithinPersistentElements = getPersistentElements().some(element =>
-        element.contains(target)
-      )
-      return !isTargetWithinPersistentElements
-    }
+    const hideOnInteractOutsideFn = useCallback(
+      (event: Event) => {
+        const target = event.target as HTMLElement
+        const isTargetWithinPersistentElements = getPersistentElements().some(element =>
+          element.contains(target)
+        )
+        return !isTargetWithinPersistentElements
+      },
+      [getPersistentElements]
+    )
 
     return (
       <Drawer
@@ -43,8 +51,9 @@ export const AssetDrawerComponent = forwardRefWithAs<AssetDrawerProps, 'div'>(
         hideOnInteractOutside={hideOnInteractOutsideFn}
         placement="bottom"
         ref={ref}
-        size={null}
         store={store}
+        // Needed to override the default height set in drawer.module.scss
+        style={{ '--drawer-height': 'calc(100% - 3rem)' } as React.CSSProperties}
         withBackdrop
         {...rest}
       >
--- a/lib/src/components/Drawer/components/Close.tsx
+++ b/lib/src/components/Drawer/components/Close.tsx
@@ -1,15 +1,17 @@
 import { DialogDismiss } from '@ariakit/react'
 
 import { CloseButton } from '@/components/CloseButton'
+import { classNames } from '@/utils'
 
+import styles from '../drawer.module.scss'
 import type { CloseProps } from '../types'
+
+const cx = classNames(styles)
 
 export const Close = ({ className }: CloseProps) => {
   return (
     <DialogDismiss
-      render={props => (
-        <CloseButton className={`absolute right-lg top-lg z-10 ${className}`} {...props} />
-      )}
+      render={props => <CloseButton className={cx('drawer-close', className)} {...props} />}
     />
   )
 }
--- a/lib/src/components/Drawer/components/Trigger.tsx
+++ b/lib/src/components/Drawer/components/Trigger.tsx
@@ -4,10 +4,8 @@ import { forwardRefWithAs } from '@/utils/forwardRefWithAs'
 
 import type { TriggerProps } from '../types'
 
-export const Trigger = forwardRefWithAs<TriggerProps, 'button'>(({ as, store, ...rest }, ref) => {
-  const Element = as || 'button'
-
-  return (
-    <DialogDisclosure ref={ref} render={props => <Element {...props} />} store={store} {...rest} />
-  )
-})
+export const Trigger = forwardRefWithAs<TriggerProps, 'button'>(
+  ({ as: As, store, ...rest }, ref) => {
+    return <DialogDisclosure ref={ref} render={As ? <As /> : undefined} store={store} {...rest} />
+  }
+)
--- a/lib/src/components/Drawer/drawer.module.scss
+++ b/lib/src/components/Drawer/drawer.module.scss
@@ -1,4 +1,4 @@
-@use '../../../src/utils/breakpoints' as breakpoints;
+@use '../../utils/breakpoints' as breakpoints;
 
 @layer components {
   .root {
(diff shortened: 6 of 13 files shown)


lib/src/components/Drawer/AssetDrawer/IconBlock.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import styles from './asset-drawer.module.scss'
+import type { IconBlockProps } from './types'
+
+const cx = classNames(styles)
+
+export const IconBlock = ({ icon: Icon, size = 'md' }: IconBlockProps) => {
+  return (
+    <div className={cx('icon-block', `size-${size}`)}>
+      <Icon color="neutral-90" size={size} />
```
cnairi: Is the color working here? I think it shouldn't 🤔 (I don't see it handled in the Icon component)

lib/src/components/Drawer/AssetDrawer/asset-drawer.module.scss
```diff
@@ -0,0 +1,99 @@
...
+    flex-shrink: 0;
+    border-radius: var(--radius-md);
+    background-color: var(--color-beige-20);
+    transition: background-color var(--timing-primary);
+
+    width: var(--icon-block-size);
+    height: var(--icon-block-size);
+
+    &.size-md {
+      --icon-block-size: 48px;
```
cnairi: It could be var(--height-48)?

lib/src/components/Drawer/AssetDrawer/asset-drawer.module.scss
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    &.size-sm {
+      --icon-block-size: 32px;
```
cnairi: It could be var(--height-32)

lib/src/components/Drawer/components/Trigger.tsx
```diff
@@ -0,0 +1,13 @@
+import { DialogDisclosure } from '@ariakit/react'
+
+import { forwardRefWithAs } from '@/utils/forwardRefWithAs'
+
+import type { TriggerProps } from '../types'
+
+export const Trigger = forwardRefWithAs<TriggerProps, 'button'>(({ as, store, ...rest }, ref) => {
+  const Element = as || 'button'
+
+  return (
```
cnairi:
```suggestion
export const Trigger = forwardRefWithAs<TriggerProps, 'button'>(
  ({ as: Element = 'button', store, ...rest }, ref) => {
    return (
```

(it's a proposal for the as syntax)

lib/src/components/Drawer/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Content } from './components/Content'
+import { Footer } from './components/Footer'
+import { Title } from './components/Title'
+import { Trigger } from './components/Trigger'
+import styles from './drawer.module.scss'
+import type { DrawerProps } from './types'
+
+const cx = classNames(styles)
+
+const DrawerComponent = forwardRefWithAs<DrawerProps, 'div'>(
```
cnairi: I don't think you need the as here? But I might be wrong

lib/src/components/Drawer/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+      withCloseButton = true,
+      ...rest
+    },
+    ref
+  ) => {
+    return (
+      <Dialog
+        backdrop={
+          withBackdrop ? (
+            <Backdrop hideOnInteractOutside={hideOnInteractOutside as boolean} />
```
cnairi, line 39: You can remove your cast here and modify your type: `export type BackdropOptions = { hideOnInteractOutside?: DialogOptions['hideOnInteractOutside'] }`

lib/src/components/Drawer/types.ts
```diff
@@ -0,0 +1,45 @@
...
+  withCloseButton?: boolean
+}
+
+export type DrawerProps = ComponentPropsWithRef<'div'> & DrawerOptions
+
+export type FooterOptions = { className?: string }
+
+export type FooterProps = ComponentPropsWithRef<'div'> & FooterOptions
+
+export type TitleOptions = { className?: string }
```
cnairi: I think you can remove all the additional className here, it's already in the div props

lib/src/components/Modal/modal.module.scss
```diff
@@ -0,0 +1,136 @@
...
+    &.hideOnInteractOutside {
+      cursor: pointer;
+    }
+
+    &[data-enter] {
+      opacity: 1;
+    }
+  }
+
+  .root {
```
cnairi, line 4: I would put the root at the top, before the backdrop, but this is optional

lib/src/components/Drawer/drawer.module.scss
```diff
@@ -0,0 +1,135 @@
...
+          }
+          &-md {
+            width: 34.375rem; /* 550px */
+          }
+          &-lg {
+            width: 42.5rem; /* 680px */
+          }
+        }
+      }
+    }
```
cnairi, line 78: From what I read, the size is not different according to the placement,
Could you just
```
    &.placement {
      &-bottom {
        bottom: 0;
        left: 0;
        right: 0;
        transform: translateY(100%);
      }
      &-left {
        bottom: 0;
        left: 0;
        top: 0;
        transform: translateX(-100%);
      }
      &-right {
        bottom: 0;
        right: 0;
        top: 0;
        transform: translateX(100%);
      }
      &-top {
        left: 0;
        right: 0;
        top: 0;
        transform: translateY(-100%);
      }
    }

     &.size {
      &-sm {
        width: 25rem; /* 400px */
      }
      &-md {
        width: 34.375rem; /* 550px */
      }
      &-lg {
        width: 42.5rem; /* 680px */
      }
    }
``` ?

File: lib/src/components/Drawer/AssetDrawer/Header.tsx [hunk 10]
```diff
@@ -0,0 +1,39 @@
...
+import { Text } from '@/components/Text'
+import { classNames } from '@/utils'
+
+import styles from './asset-drawer.module.scss'
+import { IconBlock } from './IconBlock'
+import type { HeaderProps } from './types'
+
+const cx = classNames(styles)
+
+export const Header = ({
```
Leiksa, line 12: We should use `forwardRefWithAs` here so we can make it easily overridable

lib/src/components/Drawer/AssetDrawer/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+      children,
+      className,
+      getPersistentElements: parentGetPersistentElements,
+      maxWidth,
+      store,
+      ...rest
+    },
+    ref
+  ) => {
+    const getPersistentElements = () =>
```
Leiksa: It would be great if the two callback are memoized as we don't know how Ariakit handle their memoization and we don't want to create extra re-render/listener

lib/src/components/Drawer/AssetDrawer/types.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import type { IconProps } from '@/components/Icon'
+
+export interface AssetDrawerOptions extends DialogOptions<'div'> {
+  maxWidth?: number | string
+}
+
+export type AssetDrawerProps = AssetDrawerOptions & ComponentPropsWithRef<'div'>
+
+export type HeaderProps = ComponentProps<'div'> & {
```
Leiksa, line 12: Not needed anymore with the `forwardRefWithAs`
```suggestion
export type HeaderProps = {
```

lib/src/components/Drawer/components/Close.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import { CloseButton } from '@/components/CloseButton'
+
+import type { CloseProps } from '../types'
+
+export const Close = ({ className }: CloseProps) => {
+  return (
+    <DialogDismiss
+      render={props => (
+        <CloseButton className={`absolute right-lg top-lg z-10 ${className}`} {...props} />
```
Leiksa: Could we use css styling here and not tailwind 🙈

Leiksa: (Also if not, you should use the `classNames` utility otherwise you will have extra space usage when there are no `className` passed as props)

lib/src/components/Drawer/components/Content.tsx
```diff
@@ -0,0 +1,12 @@
+import { forwardRef } from 'react'
+
+import { classNames } from '@/utils'
+
+import styles from '../drawer.module.scss'
+import type { ContentProps } from '../types'
+
+const cx = classNames(styles)
+
+export const Content = forwardRef<HTMLDivElement, ContentProps>(({ className, ...rest }, ref) => {
```
Leiksa, line 10:
```suggestion
export const Content = forwardRefWithAs<ContentProps, 'div'>(({ className, ...rest }, ref) => {
```

lib/src/components/Drawer/components/Footer.tsx
```diff
@@ -0,0 +1,12 @@
+import { forwardRef } from 'react'
+
+import { classNames } from '@/utils'
+
+import styles from '../drawer.module.scss'
+import type { FooterProps } from '../types'
+
+const cx = classNames(styles)
+
+export const Footer = forwardRef<HTMLDivElement, FooterProps>(({ className, ...rest }, ref) => {
```
Leiksa, line 10:
```suggestion
export const Footer = forwardRefWithAs<FooterProps, 'div'>(({ className, ...rest }, ref) => {
```

lib/src/components/Drawer/components/Title.tsx
```diff
@@ -0,0 +1,17 @@
+import { Text } from '@/components/Text'
+import { classNames } from '@/utils'
+
+import styles from '../drawer.module.scss'
+import type { TitleProps } from '../types'
+
+const cx = classNames(styles)
+
+export const Title = ({ children, className, ...rest }: TitleProps) => {
```
Leiksa, line 9: `forwardRefWithAs` 🙈

lib/src/components/Drawer/components/Trigger.tsx
```diff
@@ -0,0 +1,13 @@
...
+    <DialogDisclosure ref={ref} render={props => <Element {...props} />} store={store} {...rest} />
```
Leiksa: For ariakit component we don't want to run the extra render function so maybe we could just not default the as and just do `as: As` and the following (Ariakit will handle the props forwarding):

```suggestion
    <DialogDisclosure ref={ref} render={As ? <As /> : undefined} store={store} {...rest} />
```

lib/src/components/Drawer/index.tsx
Leiksa: We need it but it is not used currently in the `render` so we must pass it

lib/src/components/Drawer/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+          ) : (
+            false
+          )
+        }
+        hideOnInteractOutside={hideOnInteractOutside}
+        modal={withBackdrop}
+        ref={ref}
```
Leiksa, line 46: We default the as to `div`:
```tsx
const Element: as || 'div'

``` 

```suggestion
        className={cx('root', `placement-${placement}`, `size-${size}`, className)}
        ref={ref}
        render={<Element />}
```

anucreative:
> I spotted some differences with our current component:

This is because I updated the docs to include <Drawer.Content /> so that it looked "nicer".

lib/src/components/Drawer/index.tsx
anucreative: `DialogOptions['hideOnInteractOutside']` is `boolean` or `function`

lib/src/components/Drawer/components/Close.tsx
anucreative: > Could we use css styling here and not tailwind 🙈

`CloseButton` is a WUI component so we want to use Tailwind to override styles (from what I understand)

lib/src/components/Drawer/components/Footer.tsx
anucreative: Do we want to use `forwardRefWithAs` everywhere ?

lib/src/components/Drawer/AssetDrawer/IconBlock.tsx
theo-mesnil: Need a className

lib/src/components/Drawer/components/Close.tsx
Leiksa: This is just my opinion, but I would prefer to use Tailwind overrides only outside of components (for example, in examples or in our project setup) and never within our actual implementation. 🤔

lib/src/components/Drawer/components/Footer.tsx
Leiksa: Again in my opinion 🙈  -> it’s a missed opportunity not to leverage this flexibility, so we should make use of it wherever possible

lib/src/components/Drawer/components/Close.tsx
cnairi: cf => https://wttj.slack.com/archives/C09CT2BSF0B/p1758034277867449 In this thread, we said it was acceptable to use tailwind to compound our components when we have no other choice, I think here we can create a class in css modules to override as it's 100% static

lib/src/components/Drawer/components/Footer.tsx
cnairi: I think @theo-mesnil has another opinion on that topic, it's to allow the as only when necessary

theo-mesnil: On some components we don't want to have an as option, like label for example :) or other component with fixed element

lib/src/components/Drawer/components/Close.tsx
Leiksa: Yep, from what I understand we should use tailwind only when we have no other choice and remain at most to scss. Is that right?

lib/src/components/Drawer/components/Footer.tsx
anucreative: For me, I don't see the need for the Footer / Content to use `as`. Alors, I'm with @theo-mesnil – only use `as` when needed.

lib/src/components/Drawer/drawer.module.scss
anucreative: The `height` / `width` changes depending on whether the `placement` is `top` / `bottom` (`width` changes) or `left` / `right` (`height` changes)

lib/src/components/Drawer/AssetDrawer/index.tsx
```diff
@@ -43,8 +51,9 @@ export const AssetDrawerComponent = forwardRefWithAs<AssetDrawerProps, 'div'>(
         hideOnInteractOutside={hideOnInteractOutsideFn}
         placement="bottom"
         ref={ref}
-        size={null}
         store={store}
+        // Needed to override the default height set in drawer.module.scss
+        style={{ '--drawer-height': 'calc(100% - 3rem)' } as React.CSSProperties}
```
Leiksa, line 56: You could override it directly in the `asset-drawer` class
```
.asset-drawer {
...css
--drawer-height: calc(100% - 3rem)
}
```

anucreative: I did that at first but it didn't work because sometimes `.drawer.size-md` styles have a higher specificity than `.asset-drawer`. 

I agree that this looks like is a bit of a hack – but it works and isn't _too_ ugly. 🤓

PR #2901, opened 2025-09-24
ci: add migration script to v9 for Stack component

theo-mesnil:


scripts/upgrade-v9.mjs
```diff
@@ -300,7 +300,7 @@ async function processBoxComponents(components, shouldReplace = false) {
       let valuesNotTransformed = []
 
       // Add default flex class for Flex components
-      if (component.componentType === 'Flex') {
+      if (component.componentType === 'Flex' || component.componentType === 'Stack') {
```
anucreative, line 303: What about `<Stack />` without `direction="row"` ? This will add a `flex` class which will lay out children horizontally rather than vertically.

theo-mesnil: Good catch i create a new PR for this case

theo-mesnil: https://github.com/WTTJ/welcome-ui/pull/2902

PR #2908, opened 2025-09-25
feat: migrate radio

StephaneRob:


lib/src/components/Radio/radio.module.scss
```diff
@@ -0,0 +1,83 @@
+@layer components {
+  .root {
+    display: flex;
+    align-items: flex-start;
+    gap: var(--spacing-sm);
+    flex-direction: row;
+
+    input {
+      margin-top: var(--spacing-xs);
+    }
```
Isaacmaamouche: Would it make sens to move this under `.input` declaration ?

StephaneRob: Yep 👍

lib/src/components/Radio/types.ts
```diff
@@ -0,0 +1,18 @@
+import type { ComponentPropsWithRef } from 'react'
+
+import type { LabelProps } from '@/components/Label/types'
+
+export interface RadioOptions {
+  dataTestId?: string
+  hint?: string
+  label?: string
+  maxWidth?: number | string
+  onChange?: (event: React.MouseEvent<HTMLLabelElement>) => void
```
Isaacmaamouche: Not sure about this but `onChange`, `onClick` & `value` are already included with  `ComponentPropsWithRef<'input'>`
Do we want them on the docs ? Any thought @theo-mesnil  ?

lib/src/theme/tokens.ts
```diff
@@ -165,6 +165,7 @@ export const primitives = {
     '--color-teal-70': '#00544a',
     '--color-teal-80': '#003d35',
     '--color-teal-90': '#00211d',
+    '--color-transparent': 'transparent',
```
Isaacmaamouche, line 168: A bit overkill imo 😄

lib/src/components/Radio/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    ref
+  ) => {
+    const handleClick = (event: React.MouseEvent<HTMLLabelElement>) => {
+      event.stopPropagation()
+      if (onClick) onClick(event)
+      if (onChange) onChange(event)
+    }
+
+    return (
+      <Label className={cx('root', className)} onClick={handleClick}>
```
Leiksa, line 36: Our Radio component shouldn't handle `label` & `hint` as we have `<Field/>` component that already handle it. (As we are doing it for Checkbox per example) 🤔

lib/src/components/Radio/types.ts
Leiksa: IMO we should only have our "custom" props in the documentation 🤔

lib/src/components/Radio/index.tsx
StephaneRob: Do you remember why it was done like this in previous version?

lib/src/theme/tokens.ts
StephaneRob: Yep, but easier to use with `var`

lib/src/components/Radio/index.tsx
Leiksa: Nope I don't have any context on this unfortunately 😕 maybe @theo-mesnil ?

theo-mesnil: We wanted to destructure all the form components so that we could apply the correct padding, spacing or other styles like variants to each one.

Leiksa: What do you mean and why it was not the case on the `Checkbox` component. Also I'm finalizing the `Field` component that will forward to the input the variants, error and a11y props so we will just need to wrap the input component with a `Field` and it will work

StephaneRob: Maybe just because each checkbox is a field, radio is a bit different, a single field for many radio. So I think we should not use field for this.

theo-mesnil: Oh no, i think is because we do not use Radio but RadioGroup

StephaneRob: @theo-mesnil @Leiksa We keep the same structure, ok for you?

lib/src/theme/tokens.ts
P1X3L: @Isaacmaamouche here is the case that it has been created for 👉 https://github.com/WTTJ/welcome-ui/pull/2908/files#diff-0b6c8cec59921cb8048a2bf7c7f62d6b2a91b3b3ab69863301878cedb6bb3e2fR54

We couldn't be able to write it this way without a var referencing `transparent`

lib/src/components/Radio/index.tsx
P1X3L: You can have a single checkbox in a form.
You never have a single radio as it should be a checkbox when it is alone.

I think it is why we had no examples on radio alone and we were only showing `RadioGroup` examples to see how we are supposed to use it

P1X3L: I think we messed up the import of Label.
Before, we were only using the types from label (https://github.com/WTTJ/welcome-ui/blob/v8.4.0/lib/src/components/Radio/index.tsx#L4).

The label itself was a new tag using only the styles from the label component and overriding them (https://github.com/WTTJ/welcome-ui/blob/v8.4.0/lib/src/components/Radio/styles.ts#L55-L59)

lib/src/theme/tokens.ts
Isaacmaamouche: the link doesn't work but I trust u

lib/src/components/Radio/index.tsx
P1X3L: The label was only here to wrap the input radio in order to have its state checked wherever you click (on the radio on the text inside the label or the hint)

cnairi (approved):

PR #2910, opened 2025-09-25
feat: migrate Swiper omg

theo-mesnil:


lib/src/components/Swiper/index.tsx
```diff
@@ -0,0 +1,334 @@
...
+    withPagination = { desktop: false, mobile: false },
+  } = options
+
+  const shouldLoop = loop || autoplay
+  const { width: viewportWidth } = useViewportSize()
+  const [currentPage, setCurrentPage] = useState(0)
+  const [showLeftArrow, setShowLeftArrow] = useState(false)
+  const [showRightArrow, setShowRightArrow] = useState(false)
+  const ref = useRef<HTMLUListElement>()
+  const { screens } = useTheme()
```
Isaacmaamouche: Shouldn't this die ?

theo-mesnil: Oopsy on it

lib/src/components/Swiper/index.tsx
```diff
@@ -0,0 +1,334 @@
...
+          } as CSSProperties
+        }
+      >
+        {slides}
+      </ul>
+      <Button
+        aria-label="Previous slide"
+        className={cx(
+          'arrow',
+          'left-[10px]',
```
Isaacmaamouche: `left-[10px]` should be in the css `.arrow` declaration

lib/src/components/Swiper/index.tsx
```diff
@@ -0,0 +1,334 @@
...
+        size={navigationSize}
+        variant={withDarkUI ? 'secondary' : 'ghost'}
+      >
+        <LeftIcon />
+      </Button>
+      <Button
+        aria-label="Next slide"
+        className={cx(
+          'arrow',
+          'right-[10px]',
```
Isaacmaamouche: Same as above -> should be in the css `.arrow` declaration

lib/src/components/Swiper/index.tsx
```diff
@@ -0,0 +1,334 @@
...
+      if (e.code === 'ArrowLeft') {
+        goPrev()
+      }
+
+      if (e.code === 'ArrowRight') {
+        goNext()
+      }
+    }
+
+    window.addEventListener('keydown', handleKeys)
```
Leiksa, line 228: We should onkeydown only within the swiper container otherwise pressing the left/right arrow will make every swiper of the page navigating

**In production:**

https://github.com/user-attachments/assets/65293b81-6732-4aa8-a21d-e55a7352d75e

Leiksa (changes requested):

lib/src/components/Swiper/swiper.module.scss
```diff
@@ -0,0 +1,185 @@
...
+      max-width: var(--slideWidthMobile);
+      flex: 1;
+
+      &:last-child {
+        margin-right: 0;
+      }
+    }
+  }
+
+  .container-mobile {
```
Isaacmaamouche: Question: should we move the .container-mobile inside .container and use a nested `&-mobile` ?

lib/src/components/Swiper/index.tsx
theo-mesnil: Nice catch #yolo, on it :D

lib/src/components/Swiper/swiper.module.scss
theo-mesnil: Yes i will, it was because of testing media queries

lib/src/components/Swiper/index.tsx
theo-mesnil: @Leiksa if i'm doing this, this will remove arrow class :o

Leiksa: You could create another class `left` & `right`  nested in the `.arrow` class:
```scss
.arrow {
  // ...
  &.left {
    left: ...
  }
  &.right {
    right: ...
  }
}
```

lib/src/components/Swiper/index.tsx
theo-mesnil: I will create another issue to avoid breaking changes here

theo-mesnil: https://linear.app/wttj/issue/WUI-225/swiper-keydown-on-all-swipers

PR #2911, opened 2025-09-25
Feat: add codegeneration for breakpoints utils 

Isaacmaamouche:
### Media queries 
This pull request: 
- moves shared SCSS utilities into a dedicated theme directory
- adds generation of breakpoint variables from design tokens.
- disables stylelint rules about media queries because it can't compute imported variables
- adds `yarn check:theme` to call `generate-theme.ts` and update theme.css

### Usage 
- Import the utils with `@use` - preferred way to import SASS files. It requires you to prefix the variables with the namespace of the file.
- Values are now derived from `token.ts` - _Single source of truth 4 ever_ 

Example: 
```scss
@use 'path/to/breakpoints' as breakpoints;

.my-class {
  font-size: 16px;
  max-width: breakpoints.$breakpoint-lg; // as a value

  @media (min-width: breakpoints.$breakpoint-md) { // as a breakpoint

      max-height: calc(2 * var(--spacing-xl) + 100vh);
    }
```


lib/src/theme/theme.css
```diff
@@ -341,7 +340,7 @@
   --letter-spacing-sm: -0.019rem; /* -0.304px */
   --letter-spacing-subtitle-md: -0.0125rem; /* -0.2px */
   --letter-spacing-subtitle-sm: -0.0125rem; /* -0.2px */
-  --letter-spacing-xs: -0.0125rem; /* -0.2px */
+   --letter-spacing-xs: -0.0125rem; /* -0.2px */
```
anucreative, line 343: 🙈

package.json
```diff
@@ -8,6 +8,7 @@
     "check:audit": "/bin/bash -c 'yarn audit --level low; [[ $? -ge 2 ]] && exit 1 || exit 0'",
     "check:audit:fix": "npm_config_yes=true npx yarn-audit-fix",
     "check:docs": "node ./scripts/check-docs.mjs",
+    "check:theme": "node --import=tsx lib/src/theme/generate-theme.ts",
```
P1X3L, line 11: if we don't need `tsx` to run the generate-theme command we might get rid of it in this pr too don't U think?

P1X3L (approved):

package.json
P1X3L: We installed it together if you remember to be able to run this generation.

Isaacmaamouche: It doesn't work without tsx 😄 
Capture d’écran 2025-09-25 à 15 43 14

PR #2912, opened 2025-09-25
chore: pass through all css module files for uniformization

P1X3L:
I've gone through all `.module.scss` files to try to unfiromize and I asked some questions where I had doubts

Also, I've gotton rid of every `@import` rules that where on `global.css` as next is already using the correct css regarding these styles.

We should still need these imports in our consuming apps, but they will be builded css imports from this library…

### DIFF
--- a/lib/src/components/Accordion/accordion.module.scss
+++ b/lib/src/components/Accordion/accordion.module.scss
@@ -12,7 +12,7 @@
 
   .icon {
     flex-shrink: 0;
-    transform: rotate3d(0, 0, 0, 0);
+    transform: rotate3d(0);
     transition: var(--duration-medium);
     padding: var(--spacing-sm);
     display: flex;
--- a/lib/src/components/Link/link.module.scss
+++ b/lib/src/components/Link/link.module.scss
@@ -54,10 +54,10 @@
 }
 
 .wui-text {
-  margin-right: calc(var(--spacing-2) * -1);
-  margin-left: calc(var(--spacing-2) * -1);
-  padding-left: var(--spacing-2);
-  padding-right: var(--spacing-2);
+  margin-right: calc(var(--spacing-xxs) * -1);
+  margin-left: calc(var(--spacing-xxs) * -1);
+  padding-left: var(--spacing-xxs);
+  padding-right: var(--spacing-xxs);
   transition:
     background var(--timing-primary),
     color var(--timing-primary);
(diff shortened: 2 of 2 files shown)


lib/src/components/Link/link.module.scss
```diff
@@ -54,22 +54,21 @@
...
 .wui-text {
-  margin-right: -2px;
-  margin-left: -2px;
-  padding-left: 2px;
-  padding-right: 2px;
-  transition: background var(--timing-primary);
+  margin-right: calc(var(--spacing-2) * -1);
+  margin-left: calc(var(--spacing-2) * -1);
+  padding-left: var(--spacing-2);
+  padding-right: var(--spacing-2);
```
cnairi: Is it --spacing-2 or --spacing-xxs?

lib/src/components/Link/link.module.scss
```diff
@@ -54,22 +54,21 @@
...
-  transition:
-    background 250ms,
-    color 250ms;
   background-image: linear-gradient(
     0deg,
     var(--backgroundImageColor),
     var(--backgroundImageColor) 100%
   );
-  background-position-y: var(--backgroundPositionY, calc(200% - 2px));
+  background-position-y: var(--backgroundPositionY, calc(200% - var(--spacing-2)));
```
cnairi: same question

lib/src/components/Accordion/accordion.module.scss
```diff
@@ -12,10 +12,9 @@
 
   .icon {
     flex-shrink: 0;
-    transform: rotate3d(0);
+    transform: rotate3d(0, 0, 0, 0);
```
cnairi: I don't have any error here with rotate3d(0)

cnairi (approved):

lib/src/components/Link/link.module.scss
P1X3L: `--spacing-xxs` is even better thanks

lib/src/components/Accordion/accordion.module.scss
P1X3L: Screenshot 2025-09-26 at 12 02 09
Looks like this for me in my neovim

P1X3L: Screenshot 2025-09-26 at 12 03 06
It seems to come from stylelintplus?

PR #2917, opened 2025-09-25
docs: remove old components on website and remove styled

theo-mesnil:


lib/src/components/Alert/index.tsx
```diff
@@ -74,7 +74,7 @@ const AlertComponent = forwardRef<HTMLDivElement, AlertProps>(
         {...rest}
       >
         {!!handleClose && (
-          <CloseButton onClick={handleClose} position="absolute" right="sm" size="xs" top="sm" />
+          <CloseButton className="absolute right-sm top-sm" onClick={handleClose} size="xs" />
```
anucreative: We should avoid using Tailwind classes in our components. This should be 

````
<CloseButton className="close-button" onClick={handleClose} size="xs" />
``` 

with the following css in the `alert.module.scss`

```
.close-button {
  position: absolute;
  right: var(--spacing-sm);
  top: var(--spacing-sm);
}
```

PR #2924, opened 2025-09-26
chore: migrate time picker

anucreative:
image


RobelTekle:
@anucreative I see some issues with the style of the scrollbar (it's black) and with the `showMonthYearPicker` option

Capture d’écran 2025-09-26 à 14 19 51
Capture d’écran 2025-09-26 à 14 19 58

https://github.com/user-attachments/assets/25250050-3a5d-4b04-98a9-8ffa8a6c939a

lib/src/components/DateTimePicker/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [value])
+
+    return (
+      <div className={cx('date-time-picker')} data-testid={dataTestId}>
+        {Children.map(children, (child: React.ReactElement, i) => {
+          const isDatePicker = i === 0
+
+          return cloneElement(child, {
+            dataTestId: isDatePicker ? `${dataTestId}-date-picker` : `${dataTestId}-time-picker`,
```
RobelTekle: I don't think this is necessary since we have access to the component by component composition. Plus if I'm passing a `dataTestId` directly to `DatePicker` or `TimePicker` this will override it.

lib/src/components/DatePicker/index.tsx
RobelTekle: we can remove `popperProps` too and let it pass by the `{...rest}`

lib/src/components/TimePicker/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+            ref={ref}
+            size={size}
+            transparent={transparent}
+          />
+        }
+        dateFormat={dateFormat}
+        onChange={handleChange}
+        placeholderText={placeholder}
+        popperContainer={CustomPopper}
+        popperProps={popperProps}
```
RobelTekle: same for this `popperProps`

lib/src/components/DateTimePickerCommon/date-time-picker.scss
```diff
@@ -0,0 +1,180 @@
+/* stylelint-disable no-descending-specificity */
```
Isaacmaamouche, line 1: no `@layer components` ?

anucreative: No, because these styles override the styles imported with react-datepicker so they need to match the same specificity (external styles have a higher specificity than `@layer` styles)

PR #2925, opened 2025-09-26
feat(RadioGroup): create new component

cnairi:
## COMPATIBILITY

## QA


lib/src/components/FieldGroup/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import fieldGroupStyles from './field-group.module.scss'
+import type { FieldGroupProps } from './types'
+
+const cx = classNames(fieldGroupStyles)
+
+export const FieldGroup = forwardRefWithAs<FieldGroupProps, 'fieldset'>(
+  ({ children, className, label, required, ...rest }, ref) => (
+    <fieldset className={cx('root', className)} ref={ref} {...rest}>
+      {label ? (
```
Leiksa: `label`(legend) is required for a11y, we can still provide a `hideLabel` prop (similar to the Field component)

lib/src/components/FieldGroup/index.tsx
```diff
@@ -0,0 +1,20 @@
+import { Label } from '@/components/Label'
+import { classNames, forwardRefWithAs } from '@/utils'
+
+import fieldGroupStyles from './field-group.module.scss'
+import type { FieldGroupProps } from './types'
+
+const cx = classNames(fieldGroupStyles)
+
+export const FieldGroup = forwardRefWithAs<FieldGroupProps, 'fieldset'>(
```
Isaacmaamouche: Why `withAs` over a normal `forwardRef`?

lib/src/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+      if (disabled) return
+
+      store.setValue(valueSelected)
+      onChange?.(valueSelected)
+    }
+
+    return (
+      <FieldGroup
+        as={AriakitRadioGroup}
+        className="mb-0"
```
Leiksa: If I remember correctly, this component isn’t “exposed”. Since it’s only used internally, why not move the mb-0 directly into its styles? you could also override this with a custom prop 🤔

Isaacmaamouche: Is the `mb-0` necessary ? FieldGroup already has `margin: 0;` in lib/src/components/FieldGroup/field-group.module.scss

lib/src/components/RadioGroup/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+        data-tesid={dataTestId}
+        disabled={disabled}
+        label={label}
+        ref={ref as unknown as Ref<HTMLDivElement>}
```
Leiksa: Do you really need to type the `ref` here? 🤔

lib/src/components/FieldGroup/index.tsx
Leiksa: Because it's used like this in the RadioGroup component:
```
<FieldGroup
        as={AriakitRadioGroup}
```

lib/src/components/RadioGroup/index.tsx
Isaacmaamouche: Also, if it's always ` as={AriakitRadioGroup}` we could move that into the FieldGroup component and we wouldn't have to cast the ref type line 49

Isaacmaamouche: And also 🙈 
the `as`prop is reaching the dom

Isaacmaamouche: Image

cnairi: Super sorry about that, as we are already working with Tailwind on ATS it's super natural to write Tailwind classes.
I completely removed it as FieldGroup already has margin: 0 in its scss [04fabe4](https://github.com/WTTJ/welcome-ui/pull/2925/commits/04fabe41c4ae7e6b457ba85ab99bd8087ed02ad8)

cnairi: - mb-0 ➡️  removed, you're right I forgot to clean it: [04fabe4](https://github.com/WTTJ/welcome-ui/pull/2925/commits/04fabe41c4ae7e6b457ba85ab99bd8087ed02ad8)
- FieldGroup is mostly used without the as props: https://github.com/search?q=org%3AWTTJ%20%3CFieldGroup&type=code, and as you highlighted it, the as was not used at all before this refactoring (that's why it was spread). I completely deleted the as
- as is no longer spread (removed)
- no need to cast the ref anymore

[1595785](https://github.com/WTTJ/welcome-ui/pull/2925/commits/159578567f6c1a03a138499906c223b28fb77807)

lib/src/components/RadioGroup/index.tsx
cnairi: Thanks for spotting it, I reworked the component and simplified it, as the "as" was not used except for here (https://github.com/search?q=org%3AWTTJ%20%3CFieldGroup&type=code), and it was not handled by the FieldGroup
[1595785](https://github.com/WTTJ/welcome-ui/pull/2925/commits/159578567f6c1a03a138499906c223b28fb77807)

lib/src/components/FieldGroup/index.tsx
cnairi: I removed it, as it wasn't used anywhere else, and not properly implemented before.

lib/src/components/FieldGroup/index.tsx
cnairi: Yes thanks for spotting: [36a701a](https://github.com/WTTJ/welcome-ui/pull/2925/commits/36a701ab1dc0ada69bc3b30156525ad493e306bd)

Leiksa (approved):

PR #2926, opened 2025-09-26
Chore: migrate Search component to tailwind [WUI-172]

Isaacmaamouche:
This PR migrates Search to tailwind
**A few thing changed**
- Search is always clearable so i removed the prop and simplified the css of icon placement and spacing
- we were not reading and using props `transparent`

It also fixes icon placement right for the Select
- Overall icon placements with the clear button and the dropdown indicator is not optimal 
- Search & Select should be reworked in that aspect with InputText as a blueprint

### DIFF
--- a/lib/src/components/Select/index.tsx
+++ b/lib/src/components/Select/index.tsx
@@ -3,7 +3,7 @@ import DownshiftImport from 'downshift'
 import { matchSorter } from 'match-sorter'
 import React, { forwardRef, Fragment, useCallback, useEffect, useMemo, useState } from 'react'
 
-import { CloseButton } from '@/components/CloseButton'
+import { CloseButton as ClearButton } from '@/components/CloseButton'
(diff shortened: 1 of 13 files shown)


lib/src/components/Select/index.tsx
```diff
@@ -345,9 +346,9 @@ export const Select = forwardRef<HTMLInputElement, SelectProps>(
 
           const DeleteIcon = (
             <div className={cx('dropdown-indicator', isOpen && 'open')}>
-              {/* TODO migrate after WUI-214/clearbutton */}
               <ClearButton
                 onClick={clearSelection as unknown as React.MouseEventHandler<HTMLButtonElement>}
+                title="Clear"
```
cnairi, line 356: I think we might want animatePresence here

cnairi (approved):

PR #2934, opened 2025-09-30
fix: migration script babel import

anucreative:
Fixes `Script failed: TypeError: traverse is not a function`


scripts/migrate-v9/process-file.mjs
```diff
@@ -1,10 +1,12 @@
 import fs from 'fs'
 
 import { parse } from '@babel/parser'
-import traverse from '@babel/traverse'
+import traverseModule from '@babel/traverse'
 
 import { parsePropsString } from './parsing.mjs'
 
+const traverse = typeof traverseModule === 'function' ? traverseModule : traverseModule.default
```
Isaacmaamouche: `getModule` here too ?

PR #2953, opened 2025-10-06
Wui 232 component button

cnairi:
## COMPATIBILITY

## QA


lib/src/components/Alert/index.tsx
```diff
@@ -41,7 +41,7 @@ const AlertComponent = forwardRef<HTMLDivElement, AlertProps>(
         if (child.type === AlertSecondaryButton) {
           return cloneElement<ButtonProps>(child, {
             size,
-            variant: withAiButton ? 'tertiary-ai' : undefined,
+            variant: withAiButton ? 'primary-ai' : undefined,
```
anucreative, line 44: I don't think the `Alert.Secondary` button should not be primary (the `Alert.Button` is primary)

lib/src/components/Button/docs/examples/ai.tsx
```diff
@@ -4,8 +4,6 @@ const Example = () => {
   return (
     <>
       <Button variant="primary-ai">primary-ai</Button>
-      <Button variant="tertiary-ai">tertiary-ai</Button>
```
anucreative, line 7: Why do we remove these from the docs ?

lib/src/components/Button/docs/examples/ai.tsx
```diff
@@ -4,8 +4,6 @@ const Example = () => {
...
-      <Button variant="ghost-ai">ghost-ai</Button>
     </>
```
anucreative, line 7: Why do we remove these from the docs ? Ah, is there only one AI button now ?

lib/src/components/Button/index.tsx
```diff
@@ -41,13 +40,7 @@ export const Button = forwardRefWithAs<ButtonProps, 'button'>(
...
         {...rest}
         accessibleWhenDisabled={accessibleWhenDisabled}
-        className={cx(
-          'root',
-          variant && `variant-${variant}`,
-          shape && `shape-${shape}`,
-          size && `size-${size}`,
-          className
-        )}
+        className={cx('root', variant && `variant-${variant}`, size && `size-${size}`, className)}
```
anucreative: `variant` and `size` will always have a value so don't really need the logical AND (`&&`)

anucreative (approved):

lib/src/components/Alert/index.tsx
cnairi: Hello Robert, it's because the only ai variant is now primary-ai:
https://www.figma.com/design/U76EtvbqkiIzjqHY27IKgy/-Belonging--Actions?node-id=1-3073&t=dhYnhdDIa6pkKozk-0

lib/src/components/Button/docs/examples/ai.tsx
cnairi: Here is the list of new variant:
primary
primary-neutral
primary-ai
primary-danger

secondary
secondary-danger

tertiary
tertiary-danger

https://www.figma.com/design/U76EtvbqkiIzjqHY27IKgy/-Belonging--Actions?node-id=1-3073&t=dhYnhdDIa6pkKozk-0

lib/src/components/Button/docs/examples/ai.tsx
cnairi: yes excatly

lib/src/components/Button/index.tsx
cnairi: Yes thanks ✅

lib/src/components/Alert/index.tsx
theo-mesnil: Maybe we can have just a variant `ai` ?

cnairi: Only in dev?
Like we keep
primary, primary-danger
secondary, secondary-danger
tertiary, tertiary-danger

neutral

ai

But we would have a different logic from design @theo-mesnil

theo-mesnil: Ok yes stay iso with design of course

PR #2966, opened 2025-10-08
chore: init rework icons

StephaneRob:


website/build-app/components/Sprite/index.tsx
```diff
@@ -0,0 +1,5 @@
+import { Sprite as BaseSprite } from '@/components/Icon'
```
anucreative: Do we need this ?

lib/src/components/Icon/index.tsx
```diff
@@ -1,305 +1,27 @@
...
+  ({ 'aria-label': ariaLabel, className, name, size = 'md', ...rest }, ref) => {
+    return (
+      <svg
+        aria-hidden={ariaLabel ? 'false' : 'true'}
+        aria-label={ariaLabel}
+        className={cx('root', `size-${size}`, className)}
+        ref={ref}
+        role={ariaLabel ? 'img' : undefined}
+        {...rest}
+      >
```
anucreative, line 20: You're happy with the `aria-xxx` @Leiksa ?

website/build-app/components/Sprite/index.tsx
StephaneRob: Not really just a test, the nextjs website layout is broken as soon as we add new element 🫠 . Will fix it!

lib/src/components/Icon/index.tsx
Leiksa: yeah I did that ahah 😄

PR #2982, opened 2025-10-15
chore: belonging label + varianticon fixes

StephaneRob:


lib/src/components/Icon/icons.ts
```diff
@@ -14,6 +14,8 @@ export const arrows = [
 ] as const
 
 export const actions = [
```
anucreative, line 16: Should keep these arrays in alphabetic order – easier to scan, no decisions to make when adding new icon. 🤓

StephaneRob: Yep we could reorder them

cnairi (commented):
Quick design fix:
Capture d’écran 2025-10-15 à 11 27 02

the asterisk should be a little bit on top of the label, I achieved this result by using `flex-start`

PR #2986, opened 2025-10-16
feat: migrate Badge component for belonging

stevenpersia:
## DESCRIPTION

⚠️ New badge here! I took the initiative to:
- replace the variant name `primary` with `brand`
- set the variant to `warm` as the default
- set the size to `md` as the default

If this is a problem, I can modify it!

## HOW TO TEST

Take a look here: https://wui-236-component-badge.preview.welcome-ui.com/components/badge

## COMPATIBILITY

- [X] Tested on Safari (desktop)
- [X] Tested on Chrome (desktop)
- [X] Tested on Firefox (desktop)

## QA

### DIFF
--- a/lib/src/components/Badge/index.tsx
+++ b/lib/src/components/Badge/index.tsx
@@ -1,5 +1,6 @@
 import { forwardRef } from 'react'
 
+import { Text } from '@/components/Text'
 import { classNames } from '@/utils'
 
 import badgeStyles from './badge.module.scss'
@@ -15,6 +16,8 @@ export const Badge = forwardRef<HTMLDivElement, BadgeProps>(
     const isNumber = Number.isInteger(children)
     const shouldUseAbbreviation = isNumber && withNumberAbbreviation && (children as number) > 99
     const text = shouldUseAbbreviation ? '99+' : children
+    const textVariant = size === 'lg' ? 'label-md-strong' : 'label-sm-strong'
+    const isBadgeSmall = size === 'sm'
 
     return (
       <div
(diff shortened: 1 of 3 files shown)


lib/src/components/Badge/index.tsx
```diff
@@ -9,33 +9,17 @@ const cx = classNames(badgeStyles)
...
     ref
   ) => {
     const isNumber = Number.isInteger(children)
-    const textLength = children.toString().length
-    const hasSingleCharacter = textLength === 1
     const shouldUseAbbreviation = isNumber && withNumberAbbreviation && (children as number) > 99
     const text = shouldUseAbbreviation ? '99+' : children
 
     return (
       <div
```
theo-mesnil, line 23: I think to avoid call font size etc on style, we want to have here a Text component with correct variant depend on size of component

theo-mesnil:
Please rebase to had correct Text component

lib/src/components/Badge/index.tsx
cnairi: I agree with Theo if it's possible it's, cleaner

Here, we have two types of Text depending on the badge variant:
label-md-strong for large badge
label-sm-strong for medium badge

It should be easy to handle.

stevenpersia: You're right! It's better like that: https://github.com/WTTJ/welcome-ui/pull/2986/commits/54059dbf48a83258ed779fed47bbdbb182f12c14. Thanks to both of you ❤️

PR #2990, opened 2025-10-17
feat: migrate Avatar for belonging

theo-mesnil:


lib/src/components/Avatar/index.tsx
```diff
@@ -1,35 +1,61 @@
...
+    // Show an user icon when there is no src and no name
+    const hasPlaceholder = !src && !name
 
     return (
       <div
         aria-label={name}
-        className={cx('root', size && `size-${size}`, color && `background-${color}`, className)}
+        className={cx(
+          'root',
+          size && `size-${size}`,
```
cnairi: size is always defined as there is a default value ➡️  `size-${size}` is enough

cnairi (commented):
Don't you want to create a script to automatically migrate avatar components in our codebase?

Else it's perfect thanks 🔥

cnairi (approved):

PR #3287, opened 2026-04-28
fix: invoke hide props if function

Isaacmaamouche:
`hideOnInteractOutside` takes either a boolean or a callback that returns a boolean.
We currently only check if the props is truthy
**Fix - Check if props `hideOnInteractOutside` is a function**
-  if it is, invoke and return its result


cnairi (approved):
you can now rebase to remove your vulnerabilities related changes

PR #3299, opened 2026-06-02
fix(checkbox): manage aria-label correclty

jonathan-herbault:
## DESCRIPTION

When passing disabled={false} explicitly to the Checkbox component, a checked checkbox renders with a white background instead of the expected yellow

Root cause
aria-disabled={false} was being rendered as the string attribute aria-disabled="false" on the DOM element. The CSS selector :not([aria-disabled]) checks for the absence of the attribute, so it failed to match even when the checkbox was not disabled, preventing the yellow background from applying.

BEFORE:

https://github.com/user-attachments/assets/4a4daa15-8b78-4000-b242-b2461d15af34

AFTER: 

https://github.com/user-attachments/assets/06bae035-d3cb-43a8-84ee-79ffed22f89a

## COMPATIBILITY

## QA

### DIFF
--- a/lib/src/components/Checkbox/checkbox.module.scss
+++ b/lib/src/components/Checkbox/checkbox.module.scss
@@ -29,7 +29,7 @@
       cursor: not-allowed;
     }
 
-    &:not([aria-disabled]):hover {
+    &:not([aria-disabled='true']):hover {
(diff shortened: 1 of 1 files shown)


RobelTekle (commented):
I think we should modify the css file, where instead of just checking for `aria-disabled` we should check for `aria-disabled='true'` to apply disabled style

RobelTekle (approved):

PR #3332, opened 2026-07-29
Fix(Swiper): ensure initial slide is displayed

Isaacmaamouche:
### Bug
Swiper's `initialIndex` option was ignored whenever the swiper mounted into a subtree that was still painting (happened to me in a Next.js app) because of internal mechanism, it would always show the first slide.

#### Timeline:
1. The init effect ran too early: On the first render, during the same frame the container first paints, the react effect issued a programmatic scroll to the target slide.
2. CSS scroll-snap undid it: The slide track css has `scroll-snap-type: x mandatory`. A scroll issued mid-first-layout isn't a settled snap point yet, so the browser's snap correction yanked it back to offset 0px.
3. That revert became the internal truth for the swiper's state: The scroll-back-to-0 fired an onScroll, the debounced handler read scrollLeft (now 0px) to derive which slide is visible, computed it as the current slide and set state `currentPage` to 0. A one-time init flag then blocked any retry, so the wrong position was permanent.

A Vite SPA escaped it only by luck: its client-only mount left layout already settled before the react effect ran, so the initial scroll landed on a valid snap point and stuck.

Two latent bugs sat in the same path: 
- the default `initialIndex` of 0 computed to slide -1 (it's a 1-based value)
- undefined `initialIndex` produced NaN, both were used into the scroll math

### Fix (thank you Claude)
**Defer the initial scroll to after the first paint:**
- It's scheduled across two animation frames instead of running during the commit, so it lands once layout is settled, and snap-back no longer fires. 
- Two frames are required because a frame scheduled from an effect still runs before the next paint. 
- The second frame also gives the viewport-size hook time to resolve, so the target page is correct even when slides-per-view varies by breakpoint.

**Ignore the swiper's own scroll once:** 
- A one-shot guard makes the scroll handler skip exactly the settle that follows the initial programmatic scroll, so a scroll correction by the css can't be computed as the current page. 
- It's aimed only at the initial scroll, and only when the target isn't the first page, so ordinary arrow/keyboard/swipe navigation, which relies on that same "scroll -> state" feedback, is untouched.

Sync state directly at init:
- Since the one-shot guard suppresses the usual feedback that does it, and we skip the redundant follow-up scroll that the state change would otherwise trigger.
- Clamp the target page into valid range, which fixes the -1 default and the NaN case and makes an out-of-range initialIndex land on the last page instead of scrolling past the end.
- Also cancel the pending frame on unmount (clean StrictMode re-mount) and cancel one pending scroll-handler call on unmount.

#### none of the unit tests reproduce the bug
jsdom has no layout, paint, or scroll-snap. Tests are here to lock the fix's mechanism against future refactors (that the scroll is deferred, and that the guard is one-shot rather than permanent).

#### Bonus fixes
Two doc examples (asset-swiper, asset-swiper-title) were also fixed: their assets weren't wrapped in Swiper.Slides, so the swiper never registered any slides and initialIndex did nothing.

#### Screen captures from marketplace-front X Next.js

Before 

https://github.com/user-attachments/assets/ae5d0fba-91ab-4ad4-87a7-18ff8db6b094

After

https://github.com/user-attachments/assets/a41828df-f946-4dbc-aff0-fc6a78b9ae38

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Swiper’s initial positioning across breakpoints and multi-page layouts.
  - Prevented redundant scrolling and scroll-snap corrections during initialization.
  - Preserved reliable navigation after the initial position is set.
  - Improved handling when the requested starting slide exceeds the available slides.

- **Examples**
  - Updated modal swiper examples to use the supported slide container structure.

- **Tests**
  - Added coverage for initial positioning, deferred scrolling, boundary handling, and subsequent user navigation.

### DIFF
--- a/lib/src/components/Modal/docs/examples/asset-swiper-title.tsx
+++ b/lib/src/components/Modal/docs/examples/asset-swiper-title.tsx
@@ -22,33 +22,38 @@ const Example = () => {
         >
           <AssetModal.Content>
             <Swiper store={swiper}>
-              <AssetModal.AssetWithTitle subtitle="THE JOB IS YOURS" title="Sapin & Rétroviseur">
-                <AssetModal.Iframe>
-                  <iframe
-                    allowFullScreen
-                    src="https://www.youtube.com/embed/uSTy3cT5hbw"
-                    title="YouTube video player"
-                  />
-                </AssetModal.Iframe>
-              </AssetModal.AssetWithTitle>
-              <AssetModal.AssetWithTitle subtitle="THE JOB IS YOURS" title="Claquette & Chaussette">
-                <AssetModal.Iframe>
-                  <iframe
-                    allowFullScreen
-                    src="https://www.youtube.com/embed/1kjATTF4v5A"
-                    title="YouTube video player"
-                  />
-                </AssetModal.Iframe>
-              </AssetModal.AssetWithTitle>
-              <AssetModal.AssetWithTitle subtitle="THE JOB IS YOURS" title="Pizza & Ananas">
-                <AssetModal.Iframe>
-                  <iframe
-                    allowFullScreen
-                    src="https://www.youtube.com/embed/YlEMLSsv_3w"
-                    title="YouTube video player"
-                  />
-                </AssetModal.Iframe>
-              </AssetModal.AssetWithTitle>
+              <Swiper.Slides>
+                <AssetModal.AssetWithTitle subtitle="THE JOB IS YOURS" title="Sapin & Rétroviseur">
+                  <AssetModal.Iframe>
+                    <iframe
+                      allowFullScreen
+                      src="https://www.youtube.com/embed/uSTy3cT5hbw"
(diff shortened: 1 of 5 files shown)


guillaumewttj (commented):
I'm ok with the behavior / functionnal part of the code ... but i think that we're at a poing where the amount of comments we have in there is a strong signal to break this component into functions / hooks.
Maybe using the principle of VM + breaking each action / reaction into a named function would help reading this component more easily ?

Isaacmaamouche:
> I'm ok with the behavior / functionnal part of the code ... but i think that we're at a poing where the amount of comments we have in there is a strong signal to break this component into functions / hooks. Maybe using the principle of VM + breaking each action / reaction into a named function would help reading this component more easily ?

@guillaumewttj PR ready for review: https://github.com/WTTJ/welcome-ui/pull/3342

cnairi (approved):

PR #3345, opened 2026-08-18
fix(tag): prevent tag from nesting button

jonathan-herbault:
## DESCRIPTION

Tag rendered its own remove <button> as a DOM child of the root element.
When `as="button"` or `as="a"` was combined with `onRemove`, the root
itself became a <button>/<a>, producing invalid nested interactive
markup (<button><button aria-label="remove tag">…</button></button>).

That specific combination now renders the clickable label and the
remove button as sibling elements inside a non-interactive wrapper
that carries the pill styling, while every other case (plain onClick,
or onRemove on the default div root) keeps rendering exactly as
before.

With as button and onclick+onremove:
image

without onremove:
image

## COMPATIBILITY

## QA

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved removable tags rendered as buttons or links to prevent nested interactive elements.
  * Fixed disabled tag interactions so clicks are properly prevented.
  * Enhanced text overflow handling and icon sizing within tags.

* **Style**
  * Added consistent focus, hover, cursor, and disabled-state styling for interactive tag elements.
  * Improved support for paragraph and nested label content within tags.

### DIFF
--- a/lib/src/components/Tag/index.tsx
+++ b/lib/src/components/Tag/index.tsx
@@ -33,6 +33,7 @@ export const Tag = forwardRefWithAs<TagOptions, 'div'>((props, ref) => {
     onRemove,
     removeButtonProps,
     size = 'lg',
+    style,
     variant = 'warm',
     ...rest
   } = props
@@ -90,7 +91,7 @@ export const Tag = forwardRefWithAs<TagOptions, 'div'>((props, ref) => {
 
   if (needsSplitLayout) {
     return (
-      <div className={rootClassName}>
+      <div className={rootClassName} style={style}>
         <Component className={cx('label')} onClick={handleClick} ref={ref} {...rest}>
           {labelContent}
         </Component>
@@ -100,7 +101,7 @@ export const Tag = forwardRefWithAs<TagOptions, 'div'>((props, ref) => {
   }
 
   return (
-    <Component className={rootClassName} onClick={handleClick} ref={ref} {...rest}>
+    <Component className={rootClassName} onClick={handleClick} ref={ref} style={style} {...rest}>
       {labelContent}
       {removeButtonEl}
     </Component>
--- a/lib/src/components/Tag/index.tsx
+++ b/lib/src/components/Tag/index.tsx
@@ -40,6 +40,11 @@ export const Tag = forwardRefWithAs<TagOptions, 'div'>((props, ref) => {
   // Determine if the tag should have a square shape (only one character and no remove action)
   const isSquare = getTextLength(children) === 1 && !onRemove
 
+  // A native button/anchor root can't legally contain the remove button, so that combination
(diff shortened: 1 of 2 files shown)


lib/src/components/Tag/index.tsx
```diff
@@ -52,36 +58,52 @@ export const Tag = forwardRefWithAs<TagOptions, 'div'>((props, ref) => {
...
+        {removeButtonEl}
+      </div>
+    )
+  }
+
+  return (
+    <Component className={rootClassName} onClick={handleClick} ref={ref} style={style} {...rest}>
+      {labelContent}
+      {removeButtonEl}
     </Component>
```
cnairi, line 106: A few comments on this file:

1. Typing: rootTag is cast as `Component as unknown as string`, which defeats type safety. Should be `Component as React.ElementType` instead.
2. Naming: `needsSplitLayout` is not super representative of the condition, `isInteractiveTagWithRemove` for instance is more explicit
3. Most important comment and blocker here: we don't support style overrides in this library, only className/CSS modules via classNames. The "style" prop needs to be removed, consumers needing to customize the container should use className instead.

lib/src/components/Tag/tag.module.scss
```diff
@@ -55,34 +55,68 @@
...
+    color: inherit;
+    font: inherit;
+    text-decoration: none;
+    cursor: pointer;
+    outline: transparent solid var(--border-width-md);
+
+    &:focus-visible {
+      outline-width: var(--button-border-width-focused);
+      outline-color: var(--button-color-border-focused);
+    }
```
cnairi: Where do you get this from?
It's weird since all the other focus have different style?

https://github.com/user-attachments/assets/6eac682b-8280-43eb-8ebd-80671225a289

lib/src/components/Tag/tag.module.scss
```diff
@@ -55,34 +55,68 @@
...
+    overflow: hidden;
+    min-width: 0;
+    gap: var(--gap);
+    background: none;
+    border: none;
+    padding: 0;
+    color: inherit;
+    font: inherit;
+    text-decoration: none;
+    cursor: pointer;
```
cnairi, line 90: Not sure the cursor pointer is needed here, since you already have a cursor inherit on your label item + .root:has(> .label) {
    cursor: pointer;
    text-decoration: none;
  }

Can you check please?

jonathan-herbault: i've checked but without it i don't have a cursor pointer.

lib/src/components/Tag/tag.module.scss
jonathan-herbault: i'll remove this. it was from another label component.

cnairi (approved):

PR #3361, opened 2026-09-17
[PLA-2878] fix(build): resolve CSS chunk association via Rollup's viteMetadata

P1X3L:
Fixes a build bug where DatePicker/TimePicker CSS overrides (today highlight, day names, calendar layout, etc.) were compiled but never actually loaded by consuming apps, plus a cascade-specificity issue uncovered while fixing it.

### Root cause
The post-build `patch-css-imports.js` script matched JS chunks to CSS assets by filename equality, but `chunkFileNames` (hashed) and `assetFileNames` (unhashed) never matched for shared chunks — so CSS for shared code (incl. the DatePicker/TimePicker overrides) was built into `dist/` but never imported by any JS, making it dead weight for consumers.

### Build fix
- Replaced the script with a Rollup/Vite plugin (`injectCssImportsPlugin`) that reads each chunk's real `viteMetadata.importedCss` instead of guessing from filenames, and fails the build if any CSS asset ends up unclaimed.
- Moved from a single `vite.config.ts` with inline plugins to `lib/vite/` (renamed from `scripts/`), one file per plugin/helper.

### CSS specificity fix
- Merged `date-time-picker.scss` into `date-time-picker.module.scss` (consistent `.module.scss` naming), wrapping vendor `react-datepicker` selectors in `:global(...)`.
- Removed the `@layer components` wrapper — its priority vs. a consumer app's own Tailwind layers was unpredictable and depended on load order.
- Several overrides had the same CSS specificity as `react-datepicker`'s own default stylesheet, so whichever loaded last silently won. Fixed by anchoring:
  - popper/calendar overrides under `.picker-popper-root`, applied to our own `CustomPopper` wrapper div
  - input-wrapper overrides under `.picker-input-root`, applied to `wrapperClassName` (react-datepicker renders the input wrapper and the popper as DOM siblings, not nested, so one anchor can't cover both).
- Fixed a copy-paste bug where `TimePicker` reused `date-picker-wrapper` instead of `time-picker-wrapper`; added the matching bridge selector in `Field/field.module.scss`.

## Summary by CodeRabbit

* **Style**
  * Updated date and time picker styling to use component-scoped styles, with the picker popper displayed inside a styled root container.
  * Time picker inputs are now left-aligned, matching date picker input alignment.
  * Updated picker wrapper and input classes to support the revised styling while retaining supplied custom classes.


lib/src/components/TimePicker/index.tsx
```diff
@@ -112,7 +112,12 @@ export const TimePicker = forwardRef<HTMLInputElement, TimePickerProps>(
         showTimeSelectOnly
         timeIntervals={timeIntervals}
         {...rest}
-        wrapperClassName={cx(rest.wrapperClassName, 'date-picker-wrapper', 'field-input')}
+        wrapperClassName={joinClassNames(
```
cnairi: Can we create a ticket so we don't forget to document this utils?
- In our `welcome-ui/components-workflow.mdx` file
- Elsewhere?
- Maybe also talk about it during tomorrow's chapter
please

cnairi (approved):

lib/src/components/TimePicker/index.tsx
P1X3L: https://linear.app/wttj/issue/PLA-3000/welcome-ui-document-choice-between-hashed-classes-vs-unhashed-classes

P1X3L: Finally, removed the joinClasses after a discussion with @Leiksa 👉 https://github.com/WTTJ/welcome-ui/pull/3361/commits/4e79fc59ff7f2ff2fa43f1927c19067861e04f00

The usage of this kind of writing is very rare, so it feels useless to add some specific conventions over a problematic so rare.

